PR #2, opened 2025-12-07
fix: allow empty strings for an exact ResourceSet

infiniteregrets:


api/src/v1/access.rs
```diff
@@ -5,6 +5,45 @@ use s2_common::types::{
     stream::{StreamName, StreamNamePrefix},
 };
 use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Clone)]
+pub struct MaybeEmpty<T>(pub Option<T>);
```
shikhar: slight preference to not wrap `Option` but have an enum with variants `{ Empty, NonEmpty(T) }`, unless it is highly convenient this way...

api/src/v1/access.rs
```diff
@@ -5,6 +5,52 @@ use s2_common::types::{
...
+
+#[derive(Debug, Clone)]
+#[cfg_attr(feature = "utoipa", derive(utoipa::ToSchema))]
+pub enum MaybeEmpty<T> {
+    Empty,
+    NonEmpty(T),
+}
+
+impl<T> MaybeEmpty<T> {
+    pub fn new(value: Option<T>) -> Self {
```
shikhar: seems unused?

PR #47, opened 2026-01-12
feat: openapi spec gen

infiniteregrets:


Cargo.toml
```diff
@@ -52,6 +52,9 @@ tokio-util = "0.7"
 tower-http = "0.6"
 tracing = "0.1"
 tracing-subscriber = "0.3"
-utoipa = "5.4"
+utoipa = { version = "5.4", features = ["time"] }
```
shikhar: Keeping feature deps in the individual crates!

Cargo.toml
```diff
@@ -52,6 +52,9 @@ tokio-util = "0.7"
...
 uuid = "1.19"
 zstd = "0.13"
+
+[patch.crates-io]
+utoipa = { git = "https://github.com/infiniteregrets/utoipa", rev = "9cd181d40ac0a50551ebe1995a4d73e8685890d6" }
```
shikhar, line 61: I wonder if this will prevent publishing to crates.io

shikhar: Hopefully not as a patch, on the dep itself is a problem for sure

infiniteregrets:
@greptile-apps review pls

lite/src/handlers/mod.rs
```diff
@@ -1,3 +1,6 @@
+#[cfg(feature = "utoipa")]
+pub mod v1;
+#[cfg(not(feature = "utoipa"))]
 mod v1;
```
shikhar: it's ok just make it pub

lite/src/handlers/v1/mod.rs
```diff
@@ -1,7 +1,13 @@
...
-mod records;
-mod streams;
+#[cfg(feature = "utoipa")]
+pub mod metrics;
+#[cfg(feature = "utoipa")]
+pub mod paths;
+pub mod records;
+pub mod streams;
+#[cfg(feature = "utoipa")]
+pub mod tokens;
```
shikhar: `access_tokens`

PR #153, opened 2026-02-03
fix(lite): validate read timestamp >= until

infiniteregrets:


lite/src/handlers/v1/records.rs
```diff
@@ -48,6 +48,38 @@ fn validate_read_until(start: ReadStart, end: ReadEnd) -> Result<(), ServiceErro
...
+        end.count = end.count.map(|c| c.saturating_sub(count));
+        end.bytes = end.bytes.map(|c| c.saturating_sub(bytes));
+    }
+    (start, end)
+}
+
+fn prepare_read(
+    start: ReadStart,
+    end: v1t::stream::ReadEnd,
+    clamp_unary: bool,
```
shikhar: don't like boolean params, use an enum (Unary / Streaming)

shikhar (approved):

PR #226, opened 2026-02-19
fix(sdk): do not retry `match_seq_num` appends with `NoSideEffects` 

sgbalogh:
While it is legal to do so, in the sense that retrying appends that specify a `match_seq_num` constraint cannot result in duplicate records, it leaves users in a more confusing state.

For instance: suppose that an append (with NoSideEffects, and which supplies a match_seq_num) hits an `unavailable` error -- i.e., one which is indefinite w/r/t durability. We don't know if the append became durable or not.

Now the SDK retries it, because it is safe to do so. Upon retry, we get a match seqnum error -- i.e., a _definite_ error w/r/t durability. But the user only sees the definite error, it is not obvious that a retry had occurred. This is a more confusing outcome than if we had just returned `unavailable`. I think this outweighs the potential upside of the retry having succeeded transparently (which it could have, had that first indefinite failed append not actually become durable).

### DIFF
--- a/sdk/src/session/append.rs
+++ b/sdk/src/session/append.rs
@@ -61,7 +61,6 @@ impl AppendSessionError {
     pub fn has_no_side_effects(&self) -> bool {
         match self {
             Self::Api(err) => err.has_no_side_effects(),
-            // AckTimeout/ServerDisconnected: server may have processed appends.
             _ => false,
         }
     }


sdk/src/types.rs
```diff
@@ -288,16 +288,13 @@ impl From<Compression> for CompressionAlgorithm {
...
     NoSideEffects,
 }
 
 impl AppendRetryPolicy {
-    pub(crate) fn is_compliant(&self, input: &AppendInput) -> bool {
-        match self {
-            Self::All => true,
-            Self::NoSideEffects => input.match_seq_num.is_some(),
-        }
+    pub(crate) fn is_compliant(&self) -> bool {
```
shikhar: should we inline and remove this fn?

sdk/src/ops.rs
```diff
@@ -340,12 +340,7 @@ impl S2Stream {
...
-        let retry_enabled = self
-            .client
-            .config
-            .retry
-            .append_retry_policy
-            .is_compliant(&input);
+        let retry_enabled = self.client.config.retry.append_retry_policy.is_compliant();
         let ack = self
             .client
             .append(&self.name, input.into(), retry_enabled)
```
shikhar: would be neater to make the arg the enum value

sgbalogh:
@greptile-ai

PR #239, opened 2026-02-24
feat: support creating resources from spec

infiniteregrets:
credits: sonnet 4.6, codex 5.3

### DIFF
--- a/.github/workflows/bump-chart-version.yml
+++ b/.github/workflows/bump-chart-version.yml
@@ -1,0 +1,41 @@
+name: Bump Chart Version
+
+on:
+  workflow_dispatch:
+
+jobs:
+  bump_chart_version:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          ref: main
+          token: ${{ secrets.RELEASE_PAT }}
+
+      - name: Bump chart patch version
+        run: |
+          CURRENT_VERSION=$(grep '^version:' charts/s2-lite/Chart.yaml | awk '{print $2}')
+          if [ -z "$CURRENT_VERSION" ]; then
+            echo "ERROR: Could not read current chart version"
+            exit 1
+          fi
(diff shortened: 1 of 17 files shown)


sdk/src/ops.rs
```diff
@@ -103,6 +104,30 @@ impl S2 {
...
     }
 
+    /// Create or reconfigure a basin.
+    ///
+    /// Creates the basin if it doesn't exist, or reconfigures it to match the provided
+    /// configuration if it does. Uses HTTP PUT semantics — always idempotent.
+    ///
+    /// Returns [`CreateOrReconfigured::Created`] with the basin info if the basin was newly
+    /// created, or [`CreateOrReconfigured::Reconfigured`] if it already existed.
+    pub async fn create_or_reconfigure_basin(
```
quettabit, line 119: WDYT abt having these two new APIs feature gated in `_hidden`? thats the approach we are taking for methods that are needed only for CLI or other internal use cases

infiniteregrets: makes sense

api/src/v1/basin.rs
```diff
@@ -116,8 +116,8 @@ impl From<types::basin::BasinState> for BasinState {
 #[derive(Debug, Clone, Serialize, Deserialize)]
 #[cfg_attr(feature = "utoipa", derive(utoipa::ToSchema))]
 pub struct CreateOrReconfigureBasinRequest {
-    /// Basin configuration.
-    pub config: Option<BasinConfig>,
+    /// Basin reconfiguration.
+    pub config: Option<BasinReconfiguration>,
     /// Basin scope.
     /// This cannot be reconfigured.
```
infiniteregrets, line 122: @quettabit do you have any objections on this? mainly the semantics are that we want create_or_reconfigure for apply and that type here makes sense..

quettabit: nah, i think this makes sense given it hasn't been exercised so far and is getting to be only now.

PR #248, opened 2026-02-25
fix: Switch to the existing RELEASE_PLZ GitHub App for signed commits

emiioan:
unsigned commits, violating the repo's verified signatures rule.
                                  Switch to the existing RELEASE_PLZ GitHub App for signed commits.


emiioan:
@greptileai

emiioan:
@greptileai

PR #288, opened 2026-03-03
fix(s2s): bound decompressed frame payload size

### LINKED ISSUE #272: [Detail Bug] S2S streaming: decompression bombs can cause unbounded memory use (Gzip/Zstd)
# Detail Bug Report

https://app.detail.dev/org_89d327b3-b883-4365-b6a3-46b6701342a9/bugs/bug_26d35874-0a5c-4c96-96e9-9b8ceb590f74

# Summary
- **Context**: The `s2s.rs` file implements a framing protocol for server-to-server streaming with support for Gzip and Zstd compression, limiting compressed frames to 2 MiB.
- **Bug**: The decompression logic in `CompressedData::decompressed()` does not enforce any limit on the decompressed payload size, allowing an attacker to send a small compressed frame that expands to an arbitrarily large size during decompression.
- **Actual vs. expected**: The code validates that compressed frames are under 2 MiB but allows unlimited decompressed output, whereas it should limit both compressed and decompressed sizes to prevent resource exhaustion.
- **Impact**: An attacker can cause memory exhaustion and denial of service by sending carefully crafted compressed payloads (decompression bombs) that fit within the 2 MiB compressed limit but expand to hundreds of megabytes or gigabytes in memory.

# Code with Bug
```rust
fn decompressed(self) -> std::io::Result<Bytes> {
    match self.compression {
        CompressionAlgorithm::None => Ok(self.payload),
        CompressionAlgorithm::Gzip => {
            let mut decoder = GzDecoder::new(&self.payload[..]);
            let mut buf = Vec::with_capacity(self.payload.len() * 2);
            decoder.read_to_end(&mut buf)?; // <-- BUG 🔴 No size limit on decompression
            Ok(Bytes::from(buf.into_boxed_slice()))
        }
        CompressionAlgorithm::Zstd => {
            let mut buf = Vec::with_capacity(self.payload.len() * 2);
            zstd::stream::copy_decode(&self.payload[..], &mut buf)?; // <-- BUG 🔴 No size limit on decompression
            Ok(Bytes::from(buf.into_boxed_slice()))
        }
    }
}
```

# Explanation
- `read_to_end()` and `zstd::stream::copy_decode()` will keep growing `buf` until the decompressor reaches EOF, with no upper bound.
- The only enforced limit is on **compressed** frame size (2 MiB), so a highly-compressible payload can fit under this limit while expanding to very large memory usage during decompression.
- This decompression is reachable from the S2S frame processing path (e.g., `api/src/v1/stream/extract.rs` calls `try_into_proto()`, which triggers `decompressed()`), so an attacker that can send S2S frames can force the server to allocate large buffers before any validation occurs.

## Codebase Inconsistency
- The protocol enforces a 2 MiB **compressed** frame limit but has no corresponding **decompressed** size limit, even though protobuf decoding and validation happen only after decompression (too late to prevent memory allocation).

# Exploit Scenario
An attacker sends a frame with a compressed payload under 2 MiB (e.g., ~100 KB of Gzip-compressed zeros) that expands to 100+ MB (or more) during decompression. The server allocates the full decompressed buffer, leading to memory exhaustion and process/host instability or crash (DoS).

# Recommended Fix
Enforce a maximum decompressed size during decompression (e.g., wrap the decoder in `take(MAX_DECOMPRESSED_BYTES + 1)` and error if the limit is exceeded) for both Gzip and Zstd paths.

# History
This bug was introduced in commit 06f7173. The commit added the S2S (server-to-server) streaming protocol with compression support (Gzip and Zstd) but used unbounded decompression without validating decompressed size.

infiniteregrets:
fixes #272


api/src/v1/stream/s2s.rs
```diff
@@ -120,18 +130,39 @@ impl CompressedData {
     }
 
     fn decompressed(self) -> std::io::Result<Bytes> {
+        // Decode at most `MAX_DECOMPRESSED_PAYLOAD_BYTES + 1` bytes
+        fn read_to_end_limited(mut reader: impl Read) -> std::io::Result<Bytes> {
+            let mut limited = reader
+                .by_ref()
+                .take((MAX_DECOMPRESSED_PAYLOAD_BYTES + 1) as u64);
+            let mut buf = Vec::with_capacity(COMPRESSION_THRESHOLD_BYTES);
```
shikhar: This used to be `payload_len * 2` -- maybe keep that but cap at max decompressed len?

infiniteregrets:
@greptile-apps review?

PR #292, opened 2026-03-04
fix: stale delete-on-empty deadlines after config changes

### LINKED ISSUE #178: lite: stale delete-on-empty deadlines after config changes
Problem
- DOE deadlines are stored as (deadline -> min_age) and stream_doe relies on the stored min_age without checking current config.
- When delete-on-empty is disabled or min_age increases, existing StreamDeleteOnEmptyDeadline keys remain. Those stale deadlines can still delete streams even though config changed.

Where
- lite/src/backend/streams.rs (create/reconfigure schedule DOE but never clear old keys)
- lite/src/backend/bgtasks/stream_doe.rs (uses stored min_age; no config check)

Impact
- Disabling DOE or increasing min_age does not prevent deletions that were scheduled under prior config.

Possible fixes
- Clear existing DOE deadline keys for the stream when delete-on-empty is disabled or min_age increases.
- Or persist a config epoch/version with the deadline value and ignore mismatches in stream_doe.

infiniteregrets:
fixes #178 #274 

this change is backwards compatible as old values of doe are still accepted with epoch being 0 for them

### DIFF
--- a/lite/src/backend/kv/mod.rs
+++ b/lite/src/backend/kv/mod.rs
@@ -94,7 +94,7 @@ pub enum Key {
     StreamRecordTimestamp(StreamId, StreamPosition),
     /// (SDOED) per-deadline-per-stream, deletable, present while pending
     /// Key: TimestampSecs StreamID
-    /// Value: MinAge seconds + delete-on-empty config epoch
+    /// Value: MinAgeSeconds DoeEpoch
     StreamDeleteOnEmptyDeadline(timestamp::TimestampSecs, StreamId),
 }
 


lite/src/backend/kv/stream_doe_deadline.rs
```diff
@@ -7,7 +7,14 @@ use super::{DeserializationError, KeyType, check_exact_size, timestamp::Timestam
 use crate::backend::stream_id::StreamId;
 
 const KEY_LEN: usize = 1 + 4 + StreamId::LEN;
-const VALUE_LEN: usize = 8;
+const VALUE_LEN_V1: usize = 8;
+const VALUE_LEN_V2: usize = 16;
+
```
infiniteregrets, line 12: @shikhar thoughts on this?

Issue #302, opened 2026-03-06
[lite][QA] s2 lite fails to start with RustFS S3 backend: reqwest::Error { kind: Builder, source: BadScheme } during ListRequest


Mrhs121:
I'm trying to use RustFS as the S3-compatible storage backend for s2 lite, but startup fails with repeated object store retry errors.

RustFS startup command
```
docker run --name rustfs_local \
  -p 9000:9000 \
  -p 9001:9001 \
  -v ./rustfs-data:/data \
  -e RUSTFS_ACCESS_KEY=rustfsadmin \
  -e RUSTFS_SECRET_KEY=rustfsadmin \
  -e RUSTFS_CONSOLE_ENABLE=true \
  rustfs/rustfs:latest /data
```

Environment variables
```
export AWS_ACCESS_KEY_ID="rustfsadmin"
export AWS_SECRET_ACCESS_KEY="rustfsadmin"
export AWS_ENDPOINT_URL_S3="http://localhost:9000"
export AWS_REGION="us-east-1"

```
s2 lite command
```
s2 lite \
  --port 7070 \
  --bucket tag \
  --path s2lite
```

s2 lite starts, detects the S3 object store bucket, and picks up the static credentials from env vars, but then fails while opening the SlateDB database. It keeps retrying S3 list operations and logs the following error:
```
2026-03-06T07:11:13.309582Z  INFO s2_lite::server: args=LiteArgs { bucket: Some("tag"), local_root: None, path: "s2lite", tls: TlsConfig { tls_self: false, tls_cert: None, tls_key: None }, port: Some(7070), no_cors: false, init_file: None }
2026-03-06T07:11:13.311037Z  INFO s2_lite::server: using s3 object store bucket="tag"
2026-03-06T07:11:13.311613Z  INFO s2_lite::server: using static credentials from env vars key_id="rustfsadmin"
2026-03-06T07:11:13.453921Z  INFO s2_lite::server: pipelining disabled
[...]

PR #303, opened 2026-03-06
fix(lite): allow http endpoints for S3-compatible object stores

Mrhs121:
fix https://github.com/s2-streamstore/s2/issues/302.
When AWS_ENDPOINT_URL_S3 starts with "http://", automatically set allow_http on the S3 builder.


Mrhs121:
Happy path works as expected.
start s2 lite server
截屏2026-03-06 17 48 34
append data
截屏2026-03-06 17 48 27
data storaged in rustfs
截屏2026-03-06 17 46 18

PR #318, opened 2026-03-11
fix: reject . and .. as access token and stream names

infiniteregrets:
also adding a pre-commit hook


infiniteregrets:
@greptile-apps review

PR #349, opened 2026-03-23
feat: request-time data encryption

infiniteregrets:


cli/src/cli.rs
```diff
@@ -464,6 +464,57 @@ pub struct AppendArgs {
...
     pub linger: humantime::Duration,
+
+    #[command(flatten)]
+    pub encryption: EncryptionArgs,
+}
+
+#[derive(Args, Debug, Clone, Default)]
+#[group(multiple = true)]
+pub struct EncryptionArgs {
+    /// Hex-encoded 32-byte encryption key. Alternatively, set S2_ENCRYPTION_KEY env var.
```
shikhar: Let's use base64 consistently instead of hex including in header, shorter (43 chars rather than 64)

cli/src/cli.rs
```diff
@@ -464,6 +464,57 @@ pub struct AppendArgs {
...
+    #[arg(
+        long,
+        conflicts_with = "encryption_key",
+        group = "encryption_key_source"
+    )]
+    pub encryption_key_file: Option<PathBuf>,
+
+    /// Encryption algorithm (default: aegis-256).
+    #[arg(long, value_enum, requires = "encryption_key_source")]
+    pub encryption_algorithm: Option<EncryptionAlgorithm>,
```
shikhar: Starting to think requiring an explicit alg is better than picking a default.

cli/src/cli.rs
```diff
@@ -464,6 +464,57 @@ pub struct AppendArgs {
...
+    )]
+    pub encryption_key: Option<String>,
+
+    /// Read decryption key from file.
+    #[arg(long = "encryption-key-file", conflicts_with = "encryption_key")]
+    pub encryption_key_file: Option<PathBuf>,
+
+    /// Attest client-side encryption.
+    #[arg(long = "encryption-attest", conflicts_with_all = ["encryption_key", "encryption_key_file"])]
+    pub encryption_attest: bool,
```
shikhar: What does this do for decryption?

infiniteregrets: slipped while copying code, updated

cli/src/cli.rs
infiniteregrets: updated

cli/src/cli.rs
infiniteregrets: updated

shikhar:
@greptileai rereview

PR #369, opened 2026-04-08
feat: enforce allowed encryption modes via stream config

infiniteregrets:


lite/src/handlers/v1/records.rs
```diff
@@ -30,6 +30,27 @@ use s2_common::{
...
+    let resolved = stream_config.merge(basin_config.default_stream_config);
+    let mode = encryption.mode();
+    if !resolved.encryption_modes.contains(&mode) {
+        return Err(ServiceError::Validation(ValidationError(format!(
+            "encryption mode '{}' is not allowed on this stream",
+            mode,
+        ))));
+    }
+    Ok(())
+}
```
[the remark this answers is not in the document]
infiniteregrets: dependent fetch

common/src/types/config.rs
```diff
@@ -100,12 +104,25 @@ pub struct DeleteOnEmptyConfig {
...
 }
 
-#[derive(Debug, Clone, Default, PartialEq, Eq)]
+#[derive(Debug, Clone, PartialEq, Eq)]
 pub struct StreamConfig {
     pub storage_class: StorageClass,
     pub retention_policy: RetentionPolicy,
     pub timestamping: TimestampingConfig,
     pub delete_on_empty: DeleteOnEmptyConfig,
+    pub encryption_modes: BTreeSet<EncryptionMode>,
```
shikhar: `BTreeSet` is overkill for such a small set, `EnumSet` would be very compact

lite/src/handlers/v1/records.rs
```diff
@@ -21,14 +21,45 @@ use s2_common::{
...
+    let stream_config = match backend
+        .get_stream_config(basin.clone(), stream.clone())
+        .await
+    {
+        Ok(config) => config,
+        Err(GetStreamConfigError::StreamNotFound(_)) => OptionalStreamConfig::default(),
+        Err(e) => return Err(e.into()),
+    };
+    let basin_config = backend.get_basin_config(basin.clone()).await?;
+    let resolved = stream_config.merge(basin_config.default_stream_config);
```
shikhar: Should not require a basin config lookup here. If missing (old stream before the field was added), just default to the full set.

api/src/v1/config.rs
```diff
@@ -269,6 +301,10 @@ pub struct StreamConfig {
     /// Delete-on-empty configuration.
     #[serde(default)]
     pub delete_on_empty: Option<DeleteOnEmptyConfig>,
+    /// Allowed encryption modes for the stream.
+    /// Defaults to `["plain"]` (plaintext only) if unspecified.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub encryption_modes: Option<Vec<EncryptionMode>>,
```
shikhar: Does this need to be `Option<Vec<..>>`, wouldn't `Vec` defaulting to empty work? It seems safe and backwards-compatible to interpret an empty list as all modes.

infiniteregrets: was just copying how we do it in other places by covering with Option<T> with None being not specified

Empty vec makes sense, will change!

lite/src/backend/streams.rs
```diff
@@ -158,9 +158,23 @@ impl Backend {
...
+                .copied()
+                .unwrap_or(s2_common::encryption::ALL_ENCRYPTION_MODES);
+            if !stream_modes.is_subset(basin_modes) {
+                return Err(s2_common::types::ValidationError(
+                    "stream encryption_modes must be a subset of basin default encryption_modes"
+                        .to_owned(),
+                )
+                .into());
+            }
+        }
```
shikhar: can this be DRY-ed, same thing repeated below

lite/src/init.rs
```diff
@@ -293,6 +293,7 @@ impl From<StreamConfigSpec> for StreamReconfiguration {
                     })
                 })
                 .map_or(Maybe::Unspecified, Maybe::Specified),
+            encryption_modes: Maybe::Unspecified,
```
shikhar: needs plumbing?

lite/src/handlers/v1/records.rs
```diff
@@ -207,6 +237,7 @@ pub async fn read(
         request,
     }: ReadArgs,
 ) -> Result<Response, ServiceError> {
+    validate_encryption_mode(&backend, &basin, &stream, request.encryption()).await?;
```
shikhar: Only validate on the append path

lite/src/handlers/v1/records.rs
```diff
@@ -389,6 +420,7 @@ pub async fn append(
         request,
     }: AppendArgs,
 ) -> Result<Response, ServiceError> {
+    validate_encryption_mode(&backend, &basin, &stream, request.encryption()).await?;
```
shikhar: Hmm, the streamer maintains the `StreamConfig` in memory, so it is unfortunate to have to look that up here. I think the actual encryption belongs in this layer, but the validation probably belongs in the streamer when sequencing the records. It can check `StoredRecord.record.algorithm`.

lite/src/init.rs
infiniteregrets: was gonna be a followup for cli + tui and this, but will fold this one in here

lite/src/backend/streamer.rs
```diff
@@ -211,6 +211,19 @@ impl Streamer {
...
+            for record in records.iter() {
+                let mode = record
+                    .parts()
+                    .record
+                    .as_ref()
+                    .into_inner()
+                    .encryption_mode();
+                if !allowed_modes.contains(mode) {
+                    return Err(AppendErrorInternal::EncryptionModeNotAllowed(mode));
+                }
```
infiniteregrets, line 224: @shikhar is this ok?

common/src/types/config.rs
```diff
@@ -30,8 +30,13 @@
 use std::time::Duration;
 
 use enum_ordinalize::Ordinalize;
+use enumset::EnumSet;
 
-use crate::maybe::Maybe;
+use crate::{encryption::EncryptionMode, maybe::Maybe};
+
+pub fn default_encryption_modes() -> EnumSet<EncryptionMode> {
+    EnumSet::only(EncryptionMode::Plain)
```
shikhar, line 38: this should be all of them no? otherwise contradicts "If empty, all encryption modes are permitted"

shikhar (changes requested):
let's future proof by creating an `EncryptionConfig` (or `EncryptionPolicy`?) stanza called `encryption`, which to start only has `allowed_modes`

lite/src/backend/streamer.rs
shikhar: I think move it into `sequenced_records` where an iteration over them is already happening

shikhar: It is kinda unnecessary to check more than 1 record in a batch because all of their modes should be identical, but feels like the right thing structurally

PR #374, opened 2026-04-09
test(lite): simplify backend integration tests and tighten read coverage

shikhar:
- consolidate shared backend test helpers for session setup, virtual-time polling, and timestamped appends
- replace repetitive low-signal read cases with focused table-driven coverage for tail, count, byte, and timestamp boundaries
- make follow-mode timing assertions deterministic under paused time while preserving coverage for lag catchup and wait-reset behavior
- keep coverage around encryption, fencing, auto-create, and append session flows while reducing boilerplate

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4092,9 +4092,9 @@ dependencies = [
 
 [[package]]
 name = "rand"
-version = "0.10.0"
+version = "0.10.1"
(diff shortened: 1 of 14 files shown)


shikhar:
@greptileai rereview

shikhar:
@greptileai rereview

PR #375, opened 2026-04-09
refactor: clarify encryption spec, mode, and format semantics

shikhar:
Rename the per-request and header-carried encryption type from `EncryptionConfig` to `EncryptionSpec` across the common, API, SDK, CLI, and lite surfaces.

Also tighten the type model around encryption semantics:
- `EncryptionMode` now represents plain-aware semantics where `Option<EncryptionAlgorithm>` was acting as a disguised mode.
- decryption mismatches now use `ModeMismatch` with `EncryptionMode` values.
- `EncryptedRecordFormat` now represents the on-disk/on-wire encrypted record framing version plus algorithm, instead of overloading `EncryptionAlgorithm` with format-id responsibilities.
- `EncryptionAlgorithm` remains the pure ciphertext algorithm type.

Also clean up encryption error and record terminology:
- `EncryptionSpecError` now has structured parse and validation variants instead of a single stringly `InvalidSpec` bucket.
- user-facing mode mismatch errors now say `record encryption mode mismatch` instead of leaking `ciphertext` terminology for the full encrypted record.
- docs and test helpers now distinguish encoded encrypted records from ciphertext slices more precisely.

Testing:
- `cargo test -p s2-common encryption --lib`
- `cargo test -p s2-lite --lib sse_read_with_plain_header_emits_error_event_and_terminates`
- `cargo test -p s2-lite --lib s2s_read_with_plain_header_returns_terminal_invalid_frame`

### DIFF
--- a/api/src/v1/stream/extract.rs
+++ b/api/src/v1/stream/extract.rs
@@ -6,7 +6,7 @@ use axum::{
 use futures::StreamExt as _;
 use http::{StatusCode, request::Parts};
 use s2_common::{
-    encryption::EncryptionConfig,
+    encryption::EncryptionSpec,
     http::{ParseableHeader, extract::HeaderRejection},
     types,
 };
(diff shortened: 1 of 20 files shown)


shikhar:
@greptileai rereview

PR #376, opened 2026-04-09
feat: enforce allowed encryption modes via stream config

infiniteregrets:


cli/src/types.rs
```diff
@@ -152,6 +152,9 @@ pub struct StreamConfig {
     #[clap(flatten)]
     /// Delete-on-empty configuration.
     pub delete_on_empty: Option<DeleteOnEmptyConfig>,
+    #[arg(long, value_delimiter = ',')]
+    /// Allowed encryption modes (comma-separated). If empty, all modes are permitted.
+    pub encryption_allowed_modes: Vec<EncryptionMode>,
```
shikhar: nest in an `EncryptionConfig` object for consistency?

infiniteregrets: added

shikhar:
wondering if we should replace empty with current set of all modes, or just leave it empty for lazily interpreting as all modes allowed

shikhar:
@greptileai redo

shikhar (approved):

PR #402, opened 2026-04-17
fix(api): return error for oversized terminal frames

tarunWebBuilder:
gh pr create --title "fix(api): return error for oversized terminal frames" --body @"
## Summary
- change S2S session frame encoding to return `io::Result<Bytes>` instead of panicking on oversized payloads
- propagate terminal-message encoding failures through `FramedMessageStream`
- add regression coverage for oversized terminal error bodies

## Testing
- `cargo fmt`
- `just fmt` could not run locally because `just` could not find the configured shell under PowerShell
- `just test` could not run locally for the same shell issue
- `cargo test -p s2-api s2s --lib` was attempted but the build ran out of disk space
- `cargo check -p s2-sdk` was attempted but the environment is missing `clang-cl.exe`
"@

### DIFF
--- a/api/src/v1/stream/s2s.rs
+++ b/api/src/v1/stream/s2s.rs
@@ -224,12 +224,15 @@ impl SessionMessage {
         )?))
     }
 
-    pub fn encode(&self) -> Bytes {
+    pub fn encode(&self) -> std::io::Result<Bytes> {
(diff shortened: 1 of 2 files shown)


shikhar:
Hmm, I think panicing is actually the right behaviour here.

PR #403, opened 2026-04-17
refactor!: replace encryption modes with stream cipher metadata and key-only headers

shikhar:
Replace per-request encryption modes with basin `stream_cipher`, immutable stream `cipher` metadata, and key-only request handling.

- move stream encryption selection to basin config and persist the resolved cipher in stream metadata
- replace `s2-encryption` / `EncryptionSpec` request handling with `s2-encryption-key`
- remove stream-level encryption config and reconfiguration surfaces from API, SDK, CLI, and init schema
- expose stream cipher in stream listings and `StreamInfo`
- update lite read/append flows and tests to the new CSEK model

Known gap:
- existing encrypted streams do not yet have a compatibility or migration path from legacy `config.encryption` metadata to the new top-level `cipher` field

Testing:
- just test -p s2-common -p s2-api -p s2-lite -p s2-sdk -p s2-cli
- cargo test -p s2-lite --lib rejected_before_stream_starts

BREAKING CHANGE: replace stream-level allowed encryption modes and the `s2-encryption` header with basin `stream_cipher` metadata and the `s2-encryption-key` header.

### DIFF
--- a/api/src/data.rs
+++ b/api/src/data.rs
@@ -101,10 +101,10 @@ pub struct S2FormatHeader {
 #[cfg_attr(feature = "utoipa", derive(utoipa::IntoParams))]
 #[cfg_attr(feature = "utoipa", into_params(parameter_in = Header))]
 pub struct S2EncryptionHeader {
-    /// Optional per-request encryption spec for append/read operations.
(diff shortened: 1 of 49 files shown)


shikhar:
@greptileai rereview

shikhar:
@greptileai rerev

PR #428, opened 2026-04-30
perf(lite): avoid writing tail position key on every stream append

### LINKED ISSUE #392: [lite] avoid writing tail position key on every stream append
SlateDB now supports reverse order iteration https://github.com/slatedb/slatedb/pull/1497

We only need the tail position now as a marker when there are no records on the stream. It need not be a part of every stream write – and thus frequently overwritten and needing compaction – but rather written when a stream is initialized / becomes empty due to the retention policy or trimming.

SBALAVIGNESH123:
## Summary

Resolves #392 — Eliminates the per-append StreamTailPosition write from the hot path by deriving the tail position on-demand via reverse iteration.

## Problem

Every stream append writes a StreamTailPosition key to SlateDB. For a stream with N appends, this creates N stale key-value pairs that must be compacted by the LSM tree, making it the single largest source of write amplification in s2-lite.

## Solution

Leverage SlateDB's reverse iteration API (`IterationOrder::Descending` from slatedb/slatedb#1497) to derive the tail position from the last `StreamRecordData` key instead of maintaining an explicit marker key.

### Changes

| File | Change |
|---|---|
| `Cargo.toml` | Bump SlateDB via `[patch.crates-io]` to include reverse iteration API |
| `streamer.rs` | **Remove** the per-append `StreamTailPosition` write from `db_submit_append()` (-5 lines) |
| `core.rs` | **Add** `derive_tail_position()` using `IterationOrder::Descending` to find the last record in O(1) |
| `core.rs` | **Update** `start_streamer()` to use `derive_tail_position()` instead of reading the explicit key |
| `stream_trim.rs` | **Write** tail position marker only when a stream becomes empty after trim (for DOE eligibility) |

### What stays the same

- `StreamTailPosition` is still written at **stream creation** (`streams.rs`) — needed as the initial empty-stream marker
- `StreamTailPosition` is still **read by DOE** (`stream_doe.rs`) for `write_timestamp_secs` — only from the marker written at creation or trim-to-empty
- `StreamTailPosition` is still **deleted** during terminal trim (`stream_trim.rs`) — cleanup path unchanged

### Impact

- **Before**: Every append writes 1 extra KV pair (20 bytes) that immediately becomes stale
- **After**: The tail position key is only written at stream creation and when trimmed to empty
- **For 1M appends**: Eliminates ~1M stale compaction entries

## SlateDB Version Note

This PR requires `IterationOrder::Descending` which was merged in slatedb/slatedb#1497 (Apr 9, 2026) but not yet released to crates.io (latest is 0.12.1). The `[patch.crates-io]` section points to the merge commit `a1ce5bc`. This should be updated to a proper version once SlateDB publishes a new release.


shikhar:
thanks @SBALAVIGNESH123, it may take me a week or two to get around to this

lite/src/backend/bgtasks/stream_trim.rs
```diff
@@ -68,6 +68,23 @@ impl Backend {
     ) -> Result<(), StorageError> {
         let has_remaining_records = self.delete_records(stream_id, trim_point).await?;
         if trim_point.end < NonZeroSeqNum::MAX && !has_remaining_records {
+            // Stream became empty due to trim — write a tail position marker so
+            // derive_tail_position() and DOE eligibility checks work correctly.
+            let tail_pos = StreamPosition {
+                seq_num: trim_point.end.get(),
+                timestamp: 0,
```
shikhar: this is wrong, needs to be timestamp of last record

lite/src/backend/bgtasks/stream_trim.rs
```diff
@@ -68,6 +68,23 @@ impl Backend {
...
+            };
+            let write_ts = kv::timestamp::TimestampSecs::now();
+            let mut wb = WriteBatch::new();
+            wb.put(
```
shikhar: this cannot be sequenced after deleting the last record, as the process could crash in between and we'd be left with no records & no/stale tail marker

shikhar:
Unfortunately there is a design flaw here. This approach is incompatible with age-based retention, since those records will be deleted by SlateDB during compaction. We could be left with missing/invalid tail position.

PR #430, opened 2026-05-04
feat: add aws:us-west-2 and aws:eu-north-1 basin scopes

cristi-:
## Summary

- Adds two new \`BasinScope\` variants (\`aws:us-west-2\`, \`aws:eu-north-1\`) alongside the existing \`aws:us-east-1\`.
- Touches every layer that mirrors the enum: internal types in \`common\`, HTTP API + \`From\` impls, public SDK + \`From\` impls, CLI clap \`ValueEnum\` + new \`--scope\` flag on \`s2 create-basin\`, and the TUI with prev/next cycling on the Region row.
- Documents the unset-scope default (\`aws:us-east-1\`) on \`CreateBasinRequest.scope\` and \`CreateOrReconfigureBasinRequest.scope\` so it surfaces in the regenerated \`openapi.json\` (utoipa picks up doc comments on schema fields). Previously the default was only documented in the SDK rustdoc.

## Notes

- Background: ties into the us-west-2 cell standup planned in s2-cloud (\`new_us_west_2_cell.md\`). \`eu-north-1\` is added now to avoid a second pass; customer-facing exposure is gated server-side.
- With \`v1alpha\` sunset in s2-specs (\`a47ed3e\`), the Rust types in \`s2/common/src/types/basin.rs\` are now the **single** source of truth for \`BasinScope\` — no proto change needed.
- \`openapi.json\` regen in s2-specs will follow after merge, per the existing manual regen flow.

## Test plan

- [x] \`just fmt\`
- [x] \`cargo check --all-targets --all-features\` clean
- [x] \`cargo clippy --workspace --all-features --all-targets -- -D warnings\` clean
- [x] \`just test\` — 470/470 pass
- [x] Verified regenerated \`openapi.json\` (via \`cargo run --bin openapi -p s2-lite --features utoipa\`) lists all 3 enum variants and includes the \"defaults to \`aws:us-east-1\`\" description on the \`scope\` field.
- [x] Manual TUI smoke: cycle Region pills with ←/→ on the create-basin form
- [ ] Manual CLI smoke: \`s2 create-basin <name> --scope aws:us-west-2\` against a sandbox

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #433, opened 2026-05-04
feat!: mark s2_sdk::BasinScope as #[non_exhaustive]

cristi-:
## Summary
Adding new variants to \`s2_sdk::types::BasinScope\` is currently semver-major because the enum is exhaustive. This bit us with #430: bumping \`s2-api 0.28.4 → 0.28.5\` added enum variants in a patch release, which broke \`s2-sdk 0.27.1\`'s exhaustive match (39 errors when s2-verification tried to compile against the new s2-api).

Marks **only** \`s2_sdk::types::BasinScope\` \`#[non_exhaustive]\` so future variant additions on the public SDK surface don't break downstream consumers.

## Scope of the change
- \`s2_sdk::types::BasinScope\`: \`#[non_exhaustive]\` ← the public API surface, this is the one that matters
- \`cli/src/types.rs\` and \`cli/src/tui/ui.rs\`: gain \`_\` arms (cli matches across the crate boundary on the now-non_exhaustive sdk enum)
- \`s2_api::v1::basin::BasinScope\` and \`s2_common::types::basin::BasinScope\`: **left exhaustive**. Marking them non_exhaustive would only force \`unreachable!()\` arms in the workspace-internal \`From\` impls (api ← common, sdk ← api) — converting a compile error into a runtime panic, which is worse. The api↔sdk semver-skew that triggered #430's downstream break is a release-coordination problem (put api/common/sdk in the same release-plz \`version_group\`), not an enum-attribute one.
- Doc tweak: \"If omitted **when creating**, defaults to \`aws:us-east-1\`\" — the default only applies at create, not reconfigure.

## Test plan
- [x] \`just fmt\`
- [x] \`cargo check --workspace --all-features\` clean
- [x] \`just test\` — 470/470 pass

### DIFF
--- a/api/src/v1/basin.rs
+++ b/api/src/v1/basin.rs
@@ -173,7 +173,7 @@ pub struct CreateOrReconfigureBasinRequest {
     /// Basin reconfiguration.
     pub config: Option<BasinReconfiguration>,
     /// Basin scope.
-    /// If omitted, defaults to `aws:us-east-1`.
+    /// If omitted when creating, defaults to `aws:us-east-1`.
     /// This cannot be reconfigured.
     pub scope: Option<BasinScope>,
 }
@@ -189,6 +189,6 @@ pub struct CreateBasinRequest {
     /// Basin configuration.
     pub config: Option<BasinConfig>,
     /// Basin scope.
-    /// If omitted, defaults to `aws:us-east-1`.
+    /// If omitted when creating, defaults to `aws:us-east-1`.
     pub scope: Option<BasinScope>,
 }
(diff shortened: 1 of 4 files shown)


shikhar (approved):

api/src/v1/basin.rs
```diff
@@ -189,6 +193,6 @@ pub struct CreateBasinRequest {
     /// Basin configuration.
     pub config: Option<BasinConfig>,
     /// Basin scope.
-    /// If omitted, defaults to `aws:us-east-1`.
+    /// If omitted when creating, defaults to `aws:us-east-1`.
```
shikhar, line 192: does not need the qualifier as it is in a `CreateBasinRequest`

PR #434, opened 2026-05-04
fix(cli): TUI panic safety and UTF-8 input handling

### LINKED ISSUE #360: [Detail Bug] CLI TUI: Terminal left in raw mode after panic
# Detail Bug Report

https://app.detail.dev/org_89d327b3-b883-4365-b6a3-46b6701342a9/bugs/bug_53eff52b-424b-48a2-9036-1f95fc8076c8

Introduced in [#120](https://github.com/s2-streamstore/s2/pull/120) by @infiniteregrets on Jan 29, 2026

# Summary
- **Context**: The TUI module in `cli/src/tui/mod.rs` manually initializes terminal state (raw mode + alternate screen) before running the interactive application.
- **Bug**: If `enable_raw_mode()` succeeds and any code between setup and cleanup panics, the cleanup code is never executed, leaving the terminal in raw mode.
- **Actual vs. expected**: Terminal is left in raw mode (no echo, no line buffering) after the program panics, requiring manual `reset` or `stty sane` to restore. Expected: terminal should be restored on all exit paths including panics.
- **Impact**: Users cannot use their shell after a TUI panic - typed characters are not echoed, special characters are not processed, backspace doesn't work.

# Code with Bug
```rust
pub async fn run() -> Result<(), CliError> {
    // Setup terminal
    enable_raw_mode().map_err(...)?;
    let mut stdout = io::stdout();
    execute!(stdout, EnterAlternateScreen).map_err(...)?;
    let backend = CrosstermBackend::new(stdout);
    let mut terminal = Terminal::new(backend).map_err(...)?;

    // Create and run app
    let app = App::new(s2);
    let result = app.run(&mut terminal).await;  // <-- BUG 🔴 panic here skips terminal restoration

    // Restore terminal - UNREACHABLE IF PANIC OCCURS ABOVE
    let mut cleanup_errors = Vec::new();
    if let Err(e) = disable_raw_mode() { ... }
    // ...
}
```

# Explanation
- `enable_raw_mode()` changes the terminal device settings (e.g., via `tcsetattr` on Unix), so the terminal state persists after the process exits unless explicitly restored.
- Cleanup (`disable_raw_mode()` and leaving the alternate screen) is performed *after* `app.run()`. If `app.run()` panics, stack unwinding skips this code path, so the terminal remains in raw mode.
- The configured panic hook (`miette::set_panic_hook()`) only formats/prints the panic; it does not restore terminal state.

## Codebase Inconsistency
- Ratatui’s `try_init()` installs a panic hook that calls `restore()` (disables raw mode + leaves alternate screen) before enabling raw mode. This demonstrates the expected panic-safe pattern for TUI terminal initialization.

# Recommended Fix
Implement an RAII guard (Drop-based) and/or a panic hook to ensure terminal restoration runs on all exit paths (success, error return, and panic). For example, guard raw mode:

```rust
struct RawModeGuard;

impl RawModeGuard {
    fn new() -> io::Result<Self> {
        enable_raw_mode()?;
        Ok(Self)
    }
}

impl Drop for RawModeGuard {
    fn drop(&mut self) {
        let _ = disable_raw_mode();
    }
}

// In run():
let _raw_guard = RawModeGuard::new()
    .map_err(|e| CliError::RecordReaderInit(format!("terminal setup: {e}")))?;
```

(Analogous guard/hook should also ensure `LeaveAlternateScreen` runs.)

# History
This bug was introduced in commit 9fa3441a. The commit added the entire TUI module as a new feature for interactive mode, and the terminal setup/cleanup logic was written without RAII guards — cleanup code was placed after `app.run()` without considering panic safety. The file has had no subsequent modifications, so the bug has existed since the TUI feature was first implemented.

### LINKED ISSUE #382: [Detail Bug] TUI: Non-ASCII input crashes text fields due to invalid UTF-8 cursor byte offsets
# Detail Bug Report

https://app.detail.dev/org_89d327b3-b883-4365-b6a3-46b6701342a9/bugs/bug_9bedbdbe-c9f2-410e-8283-2a751affb936

Introduced in [#141](https://github.com/s2-streamstore/s2/pull/141) by @infiniteregrets on Jan 30, 2026

# Summary
- **Context**: The TUI text input system uses a cursor position treated as a byte index but incremented by 1 for each character typed, regardless of actual byte length.
- **Bug**: After typing a multi-byte UTF-8 character, the cursor points into the middle of that character. This causes panics at `String::insert()` (on second keypress) or at slicing operations (on render).
- **Actual vs. expected**: When a user types multi-byte characters, the application should handle them correctly. Instead, it panics with "byte index X is not a char boundary".
- **Impact**: Application crash on any non-ASCII input into affected fields.

# Code with Bug
```rust
KeyCode::Char(c) => {
    state.access_token.insert(state.cursor, c);
    state.cursor += 1;  // <-- BUG 🔴 cursor tracked as bytes but advanced by 1 char, can land mid-UTF-8 sequence
}
```

```rust
KeyCode::Left => {
    state.cursor = state.cursor.saturating_sub(1);  // <-- BUG 🔴 moves by bytes, can land mid-character
}
```

```rust
let before = &display[..cursor_in_display];  // <-- BUG 🔴 slices at arbitrary byte offset; panics if not a char boundary
let after = &display[cursor_in_display..];
```

# Explanation
- `state.cursor` is used as a **byte index** into Rust `String`s (passed to `String::insert(pos, char)` and used for `&s[..pos]` / `&s[pos..]`). These operations require `pos` to be on a UTF-8 character boundary.
- The input handlers increment/decrement the cursor by `1` per keypress, which is only correct for ASCII. For multi-byte characters (e.g., `ä`, 2 bytes), inserting the character at cursor 0 succeeds, then `cursor` becomes 1 (inside the UTF-8 sequence). Subsequent insertions or render-time slicing at cursor 1 panic with `byte index 1 is not a char boundary`.
- Cursor left/right is also byte-based (`±1`), so even if insertion is corrected, navigation can still move the cursor to invalid boundaries unless it jumps between character boundaries.

## Codebase Inconsistency
The UI already contains UTF-8 boundary-aware truncation, showing awareness that slicing must use char boundaries:
```rust
while end > 0 && !s.is_char_boundary(end) {
    end -= 1;
}
```
But cursor-based slicing/insertion lacks any equivalent boundary enforcement.

# Recommended Fix
Track cursor movement in UTF-8 **byte offsets** correctly:
- On insert: `cursor += c.len_utf8()`.
- On left/right: move to previous/next character boundary by scanning the string around the cursor (acceptable for short input fields).

# History
This bug was introduced in commit 525b9e9. The commit added cursor support to text input fields, changing from `push(c)` (append-only) to `insert(cursor, c)` with cursor tracking, but incremented the cursor by 1 byte regardless of the inserted character’s UTF-8 length.

alityb:
## Summary
- Fix terminal left in raw mode after panic (#360) via RAII `TerminalGuard`
- Fix non-ASCII input crashing text fields (#382) via UTF-8-aware cursor tracking
- Replace all 50+ raw cursor operations across TUI with shared `text_input` module

Closes #360
Closes #382

## What changed
- **`cli/src/tui/text_input.rs`** (new) — `TextInput` struct for owned fields, free functions for the shared-cursor `InputMode` pattern, proptest fuzz
- **`cli/src/tui/mod.rs`** — `TerminalGuard` with `Drop` impl, replaces manual cleanup
- **`cli/src/tui/app.rs`** — All cursor ops replaced, `SetupState`/`SettingsState` use `TextInput`
- **`cli/src/tui/ui.rs`** — Rendering uses safe split helpers instead of raw `&value[..cursor]`

## Test plan
- [x] `cargo clippy -p s2-cli -- -D warnings` clean
- [x] `cargo +nightly fmt --check` clean
- [x] 23 unit tests pass (10 text_input + 1 proptest fuzz + 3 handler integration + 9 CLI)
- [ ] Launch TUI (`cargo run -p s2-cli -- --interactive`), type/navigate/delete in token field, quit — terminal normal


shikhar:
@claude review

infiniteregrets (approved):

infiniteregrets:
we have `panic = abort` so i am not sure if the first issue is addressed

PR #437, opened 2026-05-10
refactor: introduce resource-specific create intents

shikhar:
## Summary

- add basin and stream create intent types under `s2_common::types` and remove the shared `CreateMode` enum
- update Lite create paths to use operation-specific payloads so create-only carries full config and create-or-reconfigure carries only reconfiguration patches
- preserve existing create idempotency metadata when reconfiguring existing resources and cover the behavior with regressions

## Tests

- `cargo check -p s2-lite --tests`
- `just fmt`
- `just test` (472/472 passed; nextest reported 1 leaky test while exiting successfully)

### DIFF
--- a/common/src/types/basin.rs
+++ b/common/src/types/basin.rs
@@ -7,7 +7,13 @@ use super::{
     ValidationError,
     strings::{NameProps, PrefixProps, StartAfterProps, StrProps},
 };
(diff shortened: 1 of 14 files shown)


shikhar:
@greptileai rereview

PR #465, opened 2026-05-18
fix: resolve lite stream config reconfigure defaults

### LINKED ISSUE #327: [Detail Bug] Streams: clearing a stream-level config override does not fall back to basin defaults in GET
# Detail Bug Report

https://app.detail.dev/org_89d327b3-b883-4365-b6a3-46b6701342a9/bugs/bug_e4997741-238d-401a-9afa-491100065dc8

# Summary
- **Context**: Stream configurations can inherit default values from basin configurations through a merge operation.
- **Bug**: When reconfiguring a stream to clear a field (setting it to `None` to use basin defaults), the field remains `None` in the stored configuration instead of being re-resolved to the basin default value.
- **Actual vs. expected**: After clearing a stream-level config override via reconfiguration, `get_stream_config` returns `None` for that field instead of the resolved basin default value.
- **Impact**: Users cannot rely on `get_stream_config` to show the effective configuration—cleared fields appear as `None` rather than showing the inherited basin default, breaking the consistency of config resolution semantics.

# Code with Bug
In `lite/src/backend/streams.rs`:
```rust
pub async fn reconfigure_stream(
    &self,
    basin: BasinName,
    stream: StreamName,
    reconfig: StreamReconfiguration,
) -> Result<OptionalStreamConfig, ReconfigureStreamError> {
    // ... transaction setup ...
    
    meta.config = meta.config.reconfigure(reconfig); // <-- BUG 🔴 No merge with basin defaults
    
    txn.put(&meta_key, kv::stream_meta::ser_value(&meta))?;
    // ... rest of function ...
}
```

Creation path resolves defaults before storing (same file):
```rust
let resolved: OptionalStreamConfig = resolved
    .merge(basin_meta.config.default_stream_config) // <-- merges with basin defaults
    .into();

let meta = kv::stream_meta::StreamMeta {
    config: resolved.clone(),
    // ...
};
```

# Explanation
`create_stream` stores a fully resolved `OptionalStreamConfig` by merging stream config with `basin_meta.config.default_stream_config`. In contrast, `reconfigure_stream` applies `reconfigure(...)` and persists the result without re-merging defaults.

If a user clears an override via `Specified(None)`, the stored stream config now contains `None` for that field. Since `get_stream_config` returns the stored config as-is (and does not merge with basin defaults), GET returns `None` rather than the effective basin default.

## Codebase Inconsistency
The codebase treats “effective config” as resolved-at-write during creation (`create_stream` merges defaults), but reconfiguration does not, creating divergent semantics for the same stream depending on whether defaults were applied at creation time vs after a clear-on-reconfigure.

# Recommended Fix
After applying `meta.config.reconfigure(reconfig)` in `reconfigure_stream`, re-fetch the basin meta and re-merge `basin_meta.config.default_stream_config` before persisting, matching `create_stream` behavior.

# History
This bug was introduced in commit f19a194. This commit implemented the initial s2-lite backend, including both `create_stream` and `reconfigure_stream` operations. While `create_stream` correctly merged stream configs with basin defaults, `reconfigure_stream` was implemented without this merge step, creating the inconsistency where clearing a stream-level setting doesn't fall back to the basin default.

shikhar:
## Summary
- resolve explicitly cleared lite stream config fields through basin defaults during reconfigure
- reject stream reconfigure while the containing basin is deletion-pending
- store and pass effective stream config as concrete `StreamConfig` in stream metadata, streamer state, and backend get/reconfigure returns
- keep `OptionalStreamConfig` on request/config-default surfaces where partial config is still meaningful
- clean up follow-on concrete-config callsites and stale optional-config wording
- add regression coverage for clearing top-level and nested stream config fields

## Notes
- Matches the s2-cloud patch endpoint behavior: apply the reconfiguration, merge basin default stream config, then write and return the effective stream config.

Fixes #327

## Tests
- `just fmt`
- `cargo check -p s2-lite`
- `cargo test -p s2-lite streamer::`
- `cargo test -p s2-lite test_reconfigure_stream_clears_fields_to_basin_defaults`
- `cargo test -p s2-lite backend::control_plane::stream::`
- `cargo test -p s2-lite backend::data_plane::mixed::test_concurrent_reconfigure_during_append`
- `cargo test -p s2-lite handlers::v1::streams`
- `just test`

### DIFF
--- a/common/src/types/config.rs
+++ b/common/src/types/config.rs
@@ -6,8 +6,8 @@
 //!   produced by merging optional configs with defaults using `merge()`.
 //!
 //! - Optional (`OptionalStreamConfig`, `OptionalTimestampingConfig`,
-//!   `OptionalDeleteOnEmptyConfig`): stored metadata, where `None` means "not set at this layer;
-//!   fall back to defaults."
+//!   `OptionalDeleteOnEmptyConfig`): partial configuration layers, where `None` means "not set at
+//!   this layer; fall back to defaults."
 //!
 //! - Reconfiguration (`StreamReconfiguration`, `TimestampingReconfiguration`,
 //!   `DeleteOnEmptyReconfiguration`): PATCH-style updates applied with `reconfigure()`.
@@ -91,6 +91,12 @@ pub struct TimestampingConfig {
 #[derive(Debug, Default, Clone, Copy, PartialEq, Eq)]
 pub struct DeleteOnEmptyConfig {
     pub min_age: Duration,
+}
+
+impl DeleteOnEmptyConfig {
+    pub fn min_age(&self) -> Option<Duration> {
+        Some(self.min_age).filter(|age| !age.is_zero())
+    }
 }
 
 #[derive(Debug, Clone, Default, PartialEq, Eq)]
--- a/lite/src/backend/bgtasks/basin_deletion.rs
+++ b/lite/src/backend/bgtasks/basin_deletion.rs
@@ -134,7 +134,7 @@ mod tests {
 
     use s2_common::types::{
         basin::BasinName,
-        config::BasinConfig,
+        config::{BasinConfig, StreamConfig},
         resources::ListLimit,
         stream::{StreamName, StreamNameStartAfter},
     };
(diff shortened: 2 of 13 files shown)


shikhar:
@greptileai rereview

PR #468, opened 2026-05-18
chore: scope as free string

sgbalogh:
Moves `scope` to a server-validated field. Also adds it to `list-basins` display output in cli.

### DIFF
--- a/api/src/v1/basin.rs
+++ b/api/src/v1/basin.rs
@@ -48,7 +48,7 @@ pub struct BasinInfo {
     /// Basin name.
     pub name: BasinName,
     /// Basin scope.
-    pub scope: Option<BasinScope>,
+    pub scope: Option<String>,
     /// Creation time in RFC 3339 format.
     #[serde(with = "time::serde::rfc3339")]
     pub created_at: OffsetDateTime,
@@ -71,7 +71,7 @@ impl From<types::basin::BasinInfo> for BasinInfo {
 
         Self {
             name,
-            scope: scope.map(Into::into),
+            scope,
             created_at,
             deleted_at,
             state: basin_state_for_deleted_at(deleted_at.as_ref()),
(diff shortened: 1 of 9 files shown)


api/src/v1/basin.rs
```diff
@@ -48,7 +48,7 @@ pub struct BasinInfo {
     /// Basin name.
     pub name: BasinName,
     /// Basin scope.
-    pub scope: Option<BasinScope>,
+    pub scope: Option<String>,
```
shikhar, line 51: wondering if worthwhile to make it a length-validated newtype on `CompactString`

quettabit:
just a reminder for `!` in PR title. also i think this is not a `chore`? more in the `refactor!` territory?

api/src/v1/basin.rs
```diff
@@ -173,9 +138,9 @@ pub struct EnsureBasinRequest {
     /// Basin configuration.
     pub config: Option<BasinConfig>,
     /// Basin scope.
-    /// If omitted when creating, defaults to `aws:us-east-1`.
+    /// If omitted when creating, uses the configured default scope.
```
shikhar, line 141: "configured default scope" seems to imply user can configure the default... hmm I guess that could become true.

or maybe "default scope for the service".

sdk/src/types.rs
```diff
@@ -861,37 +861,8 @@ impl From<BasinConfig> for api::config::BasinConfig {
...
-impl From<BasinScope> for api::basin::BasinScope {
-    fn from(value: BasinScope) -> Self {
-        match value {
-            BasinScope::AwsUsEast1 => api::basin::BasinScope::AwsUsEast1,
-            BasinScope::AwsUsWest2 => api::basin::BasinScope::AwsUsWest2,
-            BasinScope::AwsEuNorth1 => api::basin::BasinScope::AwsEuNorth1,
-        }
-    }
-}
+pub type BasinScope = String;
```
quettabit, line 865: do we really need this alias? i.e. wondering why not just expose as `String`

api/src/v1/basin.rs
sgbalogh: yeah good idea

PR #471, opened 2026-05-19
feat(sdk): expose `ensure_*` ops 

quettabit:
in addition to publicly exposing `ensure_*` ops from `s2-sdk` w/o any code smells, this PR folds some refactors that addresses the following smells:
- `s2-cli` directly depending on `s2-api`
    - relatedly, `ErrorCode::{Authn, Authz}` variants from `s2-api` are not used by lite and cloud for emitting error codes, so removing em and fixing the CLI usage appropriately.   
-  resource spec related types are shared between `s2-cli` and `s2-lite`, but they being homed in `s2-lite` is a minor smell (e.g. `s2 apply` is applicable not only to lite but to cloud as well)


quettabit:
@greptileai review pls

sdk/src/ops.rs
```diff
@@ -323,25 +319,22 @@ impl S2Basin {
 
     /// Ensure a stream.
     ///
-    /// Creates the stream if it doesn't exist, or ensures its config exactly matches the provided
-    /// configuration after basin defaults and global defaults are applied. Uses HTTP PUT semantics
-    /// and is always idempotent.
+    /// If the stream doesn't exist, creates the stream with specified configuration.
```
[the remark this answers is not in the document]
quettabit: fixed already.

cli/src/error.rs
```diff
@@ -166,18 +165,13 @@ impl std::fmt::Display for TokenSource {
 
 fn is_auth_error(err: &S2Error) -> bool {
     match err {
-        S2Error::Server(response) => is_auth_error_code(&response.code),
+        S2Error::Server(response) => {
+            matches!(response.code.as_str(), "authn" | "permission_denied")
+        }
```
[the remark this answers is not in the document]
quettabit: > the PR description states the server never emits it.

sorry, i meant as it never emits using those variants. it does emit via hardcoded string. so `authn` needs to exist in this match.

quettabit:
@greptileai addressed your comments. pls review again and update your score.

cli/src/error.rs
```diff
@@ -166,18 +165,13 @@ impl std::fmt::Display for TokenSource {
 
 fn is_auth_error(err: &S2Error) -> bool {
     match err {
-        S2Error::Server(response) => is_auth_error_code(&response.code),
+        S2Error::Server(response) => {
+            matches!(response.code.as_str(), "authn" | "permission_denied")
```
shikhar, line 169: sounded like "authn" is not used?

shikhar: i see your reply below

common/src/resource_spec.rs
```diff
@@ -0,0 +1,586 @@
...
+        basin::BasinName,
+        config::{
+            BasinConfig, OptionalDeleteOnEmptyConfig, OptionalStreamConfig,
+            OptionalTimestampingConfig, RetentionPolicy, StorageClass, TimestampingMode,
+        },
+        stream::StreamName,
+    },
+};
+
+#[derive(Debug, Deserialize, Default, schemars::JsonSchema)]
```
shikhar, line 19: I don't know if worth it but `schemars` dep could be put behind a non-default feature

sdk/src/types.rs
```diff
@@ -905,7 +903,7 @@ pub struct CreateBasinInput {
     pub config: Option<BasinConfig>,
     /// Scope of the basin.
     ///
-    /// Defaults to [`AwsUsEast1`](BasinScope::AwsUsEast1).
+    /// Defaults to [`AwsUsEast1`](BasinScope::AwsUsEast1). Cannot be changed once set.
```
shikhar: once created?

shikhar (approved):

cli/src/error.rs
quettabit: https://github.com/s2-streamstore/s2/pull/471#discussion_r3268577031

common/src/resource_spec.rs
quettabit: ya did consider that pattern. din wanna step into that feature gating right away. worth considering as things evolve.

shikhar: hmm i'd lean to doing to avoid new transitive dep for SDK

quettabit: ya thats the reasoning for which i considered as well, just din wanna fold that in this. but cool, can fold.

quettabit: btw, i was originally considering feature gating the whole `resource_spec` mod, but existing cases are mostly just gating specific deps at call sites. WDYT abt just gating that mod?

PR #473, opened 2026-05-19
fix(sdk): panic on invariant checks involving acks from server 

### LINKED ISSUE #444: [Detail Bug] SDK: Append session panics on malformed server acknowledgements, causing generic "session dropped" errors
# Detail Bug Report

https://app.detail.dev/org_89d327b3-b883-4365-b6a3-46b6701342a9/bugs/bug_2b35cdbf-333a-4607-a584-8e25a29688b2

# Summary
- **Context**: `process_ack` validates server acknowledgment data using assertions that panic on mismatch
- **Bug**: The assertions at lines 755-772 validate server response data with `panic!` instead of returning `Result`
- **Actual vs. expected**: Users receive `SessionDropped` instead of a descriptive error, losing diagnostic information
- **Impact**: Users receive `SessionDropped` instead of a descriptive error, losing diagnostic information

# Code with Bug
```rust
// sdk/src/session/append.rs:750-772

fn process_ack(
    ack: AppendAck,
    state: &mut SessionState,
    timer: Pin<&mut MuxTimer<N_TIMER_VARIANTS>>,
    ack_timeout: Duration,
) {
    let corresponding_append = state
        .inflight_appends
        .pop_front()
        .expect("corresponding append should be present for an ack");  // (A) Client bookkeeping

    assert!(                                                   // <-- BUG 🔴 panics on malformed server data
        ack.end.seq_num >= ack.start.seq_num,
        "ack end seq_num should be greater than or equal to start seq_num"
    );

    if let Some(end) = state.prev_ack_end {
        assert!(
            ack.end.seq_num > end.seq_num,
            "ack end seq_num should be greater than previous ack end"
        );
    }

    let num_acked_records = (ack.end.seq_num - ack.start.seq_num) as usize;
    assert_eq!(
        num_acked_records,
        corresponding_append.input.records.len(),
        "ack record count should match submitted batch size"
    );
    // ...
}
```

# Explanation
- `process_ack` mixes two kinds of validation:
  - Client bookkeeping (`expect("corresponding append should be present")`) where panic is appropriate (client state machine bug).
  - Validation of *server-provided* `AppendAck` fields (`assert!`/`assert_eq!`) where panic is inappropriate because it treats unexpected/malformed server responses as an unrecoverable client bug.
- When these assertions fire, the spawned session task terminates immediately. This bypasses the session’s error propagation logic (setting `terminal_err` and sending `Err(...)` on all inflight `ack_tx` channels). As a result, callers observe a generic `S2Error::Client("session dropped")` instead of a descriptive protocol/ack error.

## Codebase Inconsistency
- The same module already treats unexpected server acknowledgements as normal errors (not panics):
  ```rust
  resend_index = resend_index
      .checked_sub(1)
      .ok_or(AppendSessionError::UnexpectedAck)?;
  ```
  This establishes an existing pattern: unexpected server responses should be represented as `Result` errors, not `panic!`.

# Recommended Fix
- Change `process_ack` to return `Result<(), AppendSessionError>` and replace the server-data assertions with explicit validation that returns a dedicated error (e.g., `AppendSessionError::MalformedAck(String)`), while keeping the `expect(...)` panic for client bookkeeping.
- Update call sites to propagate the error with `?`.

quettabit:
closes #444 

also, moves us towards "panic on client-only invariant checks, and err if it involves server"

### DIFF
--- a/sdk/src/session/append.rs
+++ b/sdk/src/session/append.rs
@@ -44,8 +44,8 @@ pub enum AppendSessionError {
     SessionClosing,
     #[error("session dropped without calling close")]
     SessionDropped,
-    #[error("unexpected append acknowledgement during resend")]
-    UnexpectedAck,
+    #[error("invalid append acknowledgement: {0}")]
+    InvalidAck(String),
 }
 
 impl AppendSessionError {
@@ -608,7 +608,7 @@ async fn run_session(
                             state,
                             timer.as_mut(),
                             ack_timeout,
-                        );
+                        )?;
                     }
                     Some(Err(err)) => {
                         return Err(err.into());
@@ -692,10 +692,12 @@ async fn resend(
                             state,
                             timer.as_mut(),
                             ack_timeout,
-                        );
-                        resend_index = resend_index
-                            .checked_sub(1)
-                            .ok_or(AppendSessionError::UnexpectedAck)?;
+                        )?;
+                        resend_index = resend_index.checked_sub(1).ok_or_else(|| {
+                            AppendSessionError::InvalidAck(
+                                "received ack without a corresponding resent append in flight".to_string(),
+                            )
+                        })?;
                     }
                     Some(Err(err)) => {
                         return Err(err.into());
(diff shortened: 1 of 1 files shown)


quettabit:
@greptileai review pls

PR #481, opened 2026-05-20
feat(sdk): support configurable rustls crypto provider

### LINKED ISSUE #367: sdk: support explicit rustls provider choice
## Summary
Today the Rust SDK relies on rustls having a process-global default crypto provider already installed before `S2::new(...)` builds its connector. That is workable, but it is not a good long-term story if users want to choose a provider explicitly.

We closed #365 for now because hard-selecting `aws-lc-rs` inside the SDK connector is a policy decision, and we may want callers to control that choice.

## Current behavior
- `s2-sdk` depends on `hyper-rustls` with `aws-lc-rs` enabled.
- The CLI and lite binaries install `aws-lc-rs` as the default rustls provider at process startup.
- SDK tests also install `aws-lc-rs` explicitly when needed.
- Plain SDK users may need to do the same in their application if rustls cannot auto-detect a provider.

## Current workaround
Applications can call this once at startup before constructing `S2`:

```rust
let _ = rustls::crypto::aws_lc_rs::default_provider().install_default();
```

That works, but it is process-global and not a great library API.

## What we should decide
Pick and implement one of these directions:

1. Keep an explicit SDK default provider and document it clearly.
2. Expose a public connector / TLS-provider override for integrators.
3. Add a first-class `S2Config` knob for provider selection if we want to support multiple compiled-in providers.

## Acceptance
- The intended provider-selection story for `s2-sdk` is explicit.
- Users do not have to discover the `install_default()` workaround by accident.
- If overrides are supported, they are public and documented.

shikhar:
## Summary
- add SDK features for rustls crypto providers, with aws-lc-rs enabled by default
- allow opt-out with default-features = false without pulling in aws-lc-rs
- build SDK TLS configs with an explicit provider instead of installing a process-global rustls provider
- document default, ring, custom, and process-global provider paths

Closes #367

## Tests
- just fmt
- cargo check -p s2-sdk --no-default-features --all-targets
- cargo check -p s2-sdk --no-default-features --features rustls-ring --all-targets
- cargo check -p s2-sdk --all-features --all-targets
- cargo test -p s2-sdk --doc
- cargo test -p s2-sdk --doc --no-default-features
- cargo test -p s2-sdk --doc --all-features
- cargo clippy -p s2-sdk --no-default-features --all-targets -- -D warnings --allow deprecated
- cargo clippy -p s2-sdk --all-features --all-targets -- -D warnings --allow deprecated
- just test

### DIFF
--- a/sdk/README.md
+++ b/sdk/README.md
@@ -61,26 +61,6 @@ The Rust SDK provides ergonomic interface and utilities to interact with the
     }
    ```
 
-## TLS crypto provider
-
-The SDK enables the `rustls-aws-lc-rs` feature by default and uses rustls's
-`aws-lc-rs` crypto provider without installing a process-global rustls provider.
-Applications can opt out of that dependency with `default-features = false`.
-They can then enable `rustls-ring`, pass a custom provider, or install a
-process-global rustls provider before constructing an SDK client:
-
-```rust
-use s2_sdk::{S2, types::S2Config};
-
-let config = S2Config::new("<YOUR_ACCESS_TOKEN>")
-    .with_rustls_aws_lc_rs_crypto_provider();
-let s2 = S2::new(config)?;
-```
-
-Use `default-features = false` plus `rustls-ring` to choose `ring` without
-pulling in `aws-lc-rs`. For custom providers, use
-`S2Config::with_rustls_crypto_provider(...)`.
-
 ## Examples
 
 The [`examples`](./examples) directory in this repository contains a variety of
--- a/sdk/README.md
+++ b/sdk/README.md
@@ -61,6 +61,26 @@ The Rust SDK provides ergonomic interface and utilities to interact with the
     }
    ```
 
+## TLS crypto provider
+
+The SDK enables the `rustls-aws-lc-rs` feature by default and uses rustls's
+`aws-lc-rs` crypto provider without installing a process-global rustls provider.
+Applications can opt out of that dependency with `default-features = false`.
+They can then enable `rustls-ring`, pass a custom provider, or install a
+process-global rustls provider before constructing an SDK client:
+
+```rust
+use s2_sdk::{S2, types::S2Config};
+
+let config = S2Config::new("<YOUR_ACCESS_TOKEN>")
+    .with_rustls_aws_lc_rs_crypto_provider();
+let s2 = S2::new(config)?;
+```
+
+Use `default-features = false` plus `rustls-ring` to choose `ring` without
+pulling in `aws-lc-rs`. For custom providers, use
+`S2Config::with_rustls_crypto_provider(...)`.
+
 ## Examples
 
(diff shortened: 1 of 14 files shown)


sdk/examples/docs_configuration.rs
```diff
@@ -52,5 +52,15 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+    // Example: Explicit rustls crypto provider
+    #[cfg(feature = "rustls-aws-lc-rs")]
+    {
+        let access_token = std::env::var("S2_ACCESS_TOKEN").unwrap_or_else(|_| "demo".into());
+        // ANCHOR: rustls-crypto-provider
+        let client = S2::new(S2Config::new(access_token).with_rustls_aws_lc_rs_crypto_provider())?;
+        // ANCHOR_END: rustls-crypto-provider
+        println!("Created client with rustls crypto provider: {:?}", client);
+    }
+
```
quettabit: oh do we need this in mintlify SDK docs page? i'm guessing this is a rust only thing?

sdk/src/lib.rs
```diff
@@ -40,6 +40,34 @@ The Rust SDK provides ergonomic wrappers and utilities to interact with the
...
+    .with_rustls_aws_lc_rs_crypto_provider();
+let s2 = S2::new(config)?;
+# }
+# Ok::<_, Box<dyn std::error::Error>>(())
+```
+
+For custom providers, use [`S2Config::with_rustls_crypto_provider`]. When no SDK
+provider feature or custom provider is configured, the default connector falls
+back to rustls's process-global provider and returns an error if one has not
+been installed.
```
quettabit: ohh is this that important of an info, which needs to end up in crates.io page?

sdk/README.md
```diff
@@ -61,6 +61,26 @@ The Rust SDK provides ergonomic interface and utilities to interact with the
...
+
+let config = S2Config::new("<YOUR_ACCESS_TOKEN>")
+    .with_rustls_aws_lc_rs_crypto_provider();
+let s2 = S2::new(config)?;
+```
+
+Use `default-features = false` plus `rustls-ring` to choose `ring` without
+pulling in `aws-lc-rs`. For custom providers, use
+`S2Config::with_rustls_crypto_provider(...)`.
+
```
quettabit: similar Q as https://github.com/s2-streamstore/s2/pull/481/changes#r3276107518, i'd avoid this pollution

sdk/src/lib.rs
quettabit: a related docs failure https://github.com/s2-streamstore/s2/actions/runs/26136706508/job/76873665002?pr=481#step:5:23

quettabit (approved):

sdk/examples/docs_configuration.rs
shikhar: yep, will remove

PR #482, opened 2026-05-20
refactor!: scope as free string

sgbalogh:


api/src/v1/basin.rs
```diff
@@ -1,6 +1,7 @@
 use s2_common::types::{
     self,
     basin::{BasinName, BasinNamePrefix, BasinNameStartAfter},
+    scope::ScopeName,
```
shikhar, line 4: I think `BasinScope` is actually a good identifier, because scoping often gets used in an authz context

shikhar: we're going for another name

PR #483, opened 2026-05-20
fix(sdk): producer returns terminal err without draining ready acks

### LINKED ISSUE #440: [Detail Bug] Producer returns terminal error without draining ready acknowledgements, causing false failures and duplicate appends
# Detail Bug Report

https://app.detail.dev/org_89d327b3-b883-4365-b6a3-46b6701342a9/bugs/bug_fa32c659-425e-40eb-9984-f975b11ede67

# Summary
- **Context**: The `Producer` in `sdk/src/producer.rs` batches records and submits them to an append session, tracking pending acknowledgements in `claimable_tickets`.
- **Bug**: When `session.submit()` fails with an error, the `run()` function returns immediately without processing `claimable_tickets`, dropping pending acknowledgements for records that have already been successfully appended.
- **Actual vs. expected**: Users receive error responses for records that were actually appended successfully, instead of receiving the correct acknowledgements.
- **Impact**: Data inconsistency - records exist on the stream but users believe they failed, potentially leading to duplicate appends on retry.

# Code with Bug
```rust
// sdk/src/producer.rs
Err(err) => {
    propagate_terminal_error(
        err,
        &terminal_err,
        &mut pending_acks,
        &mut stashed_submission,
        &mut close_tx,
        &mut cmd_rx,
    )
    .await;
    return;  // <-- BUG 🔴 claimable_tickets is dropped without dispatching already-ready Ok(ack)
}
```

# Explanation
`claimable_tickets` holds futures that resolve when the session acknowledges a submitted batch. A batch can be fully acknowledged (`Ok(ack)`) and therefore *ready in `claimable_tickets`*, but if a subsequent `session.submit()` fails (e.g., session terminates due to a non-retryable error), the producer takes the error branch and returns immediately. Because `tokio::select!` picks a pseudo-random ready branch when multiple are ready, this can happen even when an earlier `claimable_tickets.next()` is already ready with `Ok(ack)`. Returning drops `claimable_tickets` without dispatching those ready acknowledgements, so user tickets resolve to an error (e.g., dropped) despite the record having been appended.

## Codebase Inconsistency
The session’s termination path only errors *still-inflight* appends; already-acknowledged batches have been removed from `inflight_appends` and are unaffected, so their `BatchSubmitTicket` can remain `Ok(ack)` even after session termination:

```rust
// sdk/src/session/append.rs
let corresponding_append = state
    .inflight_appends
    .pop_front()  // <-- removed from inflight on ack
    .expect("corresponding append should be present for an ack");
let _ = corresponding_append.ack_tx.send(Ok(ack));
```

```rust
// sdk/src/session/append.rs
for inflight_append in state.inflight_appends.drain(..) {  // drains only still-inflight
    let _ = inflight_append.ack_tx.send(Err(err.clone()));
}
```

# Recommended Fix
Before returning on `session.submit()` error (or inside `propagate_terminal_error`), drain *ready* items from `claimable_tickets` and call `dispatch_acks(...)` for them, so already-acknowledged batches are delivered to users even when a later submission fails.

quettabit:
closes #440

### DIFF
--- a/sdk/src/producer.rs
+++ b/sdk/src/producer.rs
@@ -241,8 +241,8 @@ impl Producer {
             match_seq_num: config.match_seq_num,
         };
 
-        let mut pending_acks: VecDeque<PendingRecordAck> = VecDeque::new();
-        let mut claimable_tickets: FuturesUnordered<_> = FuturesUnordered::new();
+        let mut pending_batch_acks = FuturesUnordered::new();
+        let mut pending_record_acks = VecDeque::new();
         let mut close_tx: Option<oneshot::Sender<Result<(), S2Error>>> = None;
         let mut stashed_submission: Option<StashedSubmission> = None;
         let mut submit_fut: Option<SubmitFuture> = None;
@@ -261,7 +261,7 @@ impl Producer {
                     let submission = stashed_submission
                         .take()
                         .expect("stashed_submission should not be None");
-                    pending_acks.push_back(PendingRecordAck {
+                    pending_record_acks.push_back(PendingRecordAck {
                         ack_tx: submission.ack_tx,
                         _permit: submission.permit,
                     });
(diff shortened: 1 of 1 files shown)


quettabit:
@greptileai review pls

PR #485, opened 2026-05-20
refactor!: rename scope -> location, treat it as a string; add related RPCs

sgbalogh:
- renames `scope` to `location`; this represents the logical placement of a basin
- `location` is now a server-validated string, not an enum
- adds RPCs for listing all accessible locations, and getting/setting the default location for an account
  - left unimplemented in `lite`
- adds location display to CLI `ls` / `list_basins`

Screenshot 2026-05-18 at 4 33 44 PM


api/src/v1/basin.rs
```diff
@@ -172,10 +138,10 @@ pub enum BasinState {
 pub struct EnsureBasinRequest {
     /// Basin configuration.
     pub config: Option<BasinConfig>,
-    /// Basin scope.
-    /// If omitted when creating, defaults to `aws:us-east-1`.
+    /// Basin location.
+    /// If omitted when creating, uses the default location for the service.
```
shikhar: I guess now with the setter/getter, this should be "for the account"? Implicitly a service-level default if that is not configured.

api/specs
shikhar, line 1: this should be left for the auto-created submodule sync PR

shikhar (approved):

api/specs
sgbalogh: (discussed in slack, going to leave, auto-created PR can be ignored)

PR #497, opened 2026-05-29
fix(cli): incorrect propagation of unspecified `delete_on_empty.min_age` via `apply`

### LINKED ISSUE #491: [Detail Bug] CLI apply: `delete_on_empty: {}` incorrectly disables delete-on-empty instead of inheriting basin defaults
# Detail Bug Report

https://app.detail.dev/org_89d327b3-b883-4365-b6a3-46b6701342a9/bugs/bug_db455d1a-2fa0-4bf0-8a59-6153c94213d1

Introduced in [#471](https://github.com/s2-streamstore/s2/pull/471) by @quettabit on May 19, 2026

# Summary
- **Context**: When a user writes:
  ```yaml
  streams:
    - name: my-stream
      config:
        delete_on_empty: {}
  ```
- **Bug**: When a user writes `delete_on_empty: {}` in their spec, the CLI's `apply` path sets the stream's delete-on-empty to disabled (min_age=0), overriding any basin default. The user's intent to "not specify" is lost.
- **Actual vs. expected**: Actual behavior is: "I am explicitly setting delete_on_empty to its default value (disabled)." Expected behavior is: "I am not specifying a delete_on_empty configuration." The system should inherit from basin defaults.
- **Impact**: Streams can be created/updated with delete-on-empty disabled even when basin defaults would enable it; dry-run output may indicate inheritance from basin defaults but apply results differ.

# Code with Bug
```rust
if let Some(delete_on_empty) = config.delete_on_empty {
    sdk_config = sdk_config.with_delete_on_empty(delete_on_empty_spec_to_sdk(delete_on_empty));
    // <-- BUG 🔴 always sends Some(delete_on_empty) to server, even for `{}`
}
```

```rust
fn delete_on_empty_spec_to_sdk(delete_on_empty: DeleteOnEmptySpec) -> s2_sdk::types::DeleteOnEmptyConfig {
    // ...
    s2_sdk::types::DeleteOnEmptyConfig { min_age_secs: 0 }
    // <-- BUG 🔴 converts min_age: None (empty object) into an explicit 0 value
}
```

# Explanation
`DeleteOnEmptySpec` uses `Option<HumanDuration>` so the spec can distinguish:
- `delete_on_empty: {}` / `min_age` absent → user did not specify a value
- `delete_on_empty: { min_age: "0s" }` → user explicitly disabled delete-on-empty

In the apply path, `delete_on_empty_spec_to_sdk` collapses the `None` case into `DeleteOnEmptyConfig { min_age_secs: 0 }`. The caller then unconditionally calls `.with_delete_on_empty(...)`, which sets `StreamConfig.delete_on_empty = Some(...)`. On the server side, this becomes `OptionalDeleteOnEmptyConfig { min_age: Some(Duration::ZERO) }`, and merge logic prefers `self.min_age` over basin defaults, so basin defaults are ignored.

## Codebase Inconsistency
The dry-run path preserves intent by converting the spec to an optional config type that keeps `min_age: None`, allowing basin defaults to apply:
```rust
OptionalDeleteOnEmptyConfig {
    min_age: s.min_age.map(|h| h.0),  // None stays None!
}
```
This makes dry_run and apply interpret the same spec differently.

# Failing Test
```rust
#[test]
fn test_delete_on_empty_empty_object_vs_explicit_zero() {
    // Case 1: Empty object should mean "not specified"
    let spec = DeleteOnEmptySpec { min_age: None };
    let sdk = delete_on_empty_spec_to_sdk(spec);
    assert!(sdk.is_none(), "empty object should produce None");

    // Case 2: Explicit zero should mean "disabled"
    let spec = DeleteOnEmptySpec { min_age: Some(HumanDuration(Duration::ZERO)) };
    let sdk = delete_on_empty_spec_to_sdk(spec);
    assert!(sdk.is_some(), "explicit zero should produce Some");
    assert_eq!(sdk.unwrap().min_age_secs, 0);
}
```

# Recommended Fix
Change `delete_on_empty_spec_to_sdk` to return `Option<DeleteOnEmptyConfig>` and update the caller to only set `delete_on_empty` when `min_age` is present (e.g., using `and_then`), so `{}` maps to `None` and inherits basin defaults.

# History
This bug was introduced in commit 8369f94. The commit refactored the CLI's apply path to use direct SDK type conversions instead of the existing `From` trait implementations. The new `delete_on_empty_spec_to_sdk` function was written to always return a `DeleteOnEmptyConfig` (with `min_age_secs: 0` when `min_age` is `None`), unlike the correct `From<DeleteOnEmptySpec> for OptionalDeleteOnEmptyConfig` implementation that preserves the `None` distinction. The bug slipped in because the new conversion function's semantics diverged from the existing `From` impl in the same codebase.

quettabit:
closes #491

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -203,7 +203,7 @@ jobs:
             "ref": "main",
             "lang": "go",
             "go-version": "1.24",
-            "test_cmd": "go test -v -count=1 -skip 'WithScope|AccessToken|Metrics|Client_InvalidToken' ./s2/..."
+            "test_cmd": "go test -v -count=1 -skip 'WithScope|AccessToken|Metrics|Client_InvalidToken|Location' ./s2/..."
           },
           {
             "name": "typescript",
@@ -219,7 +219,7 @@ jobs:
             "ref": "main",
             "lang": "python",
             "uv-version": "0.11.6",
-            "test_cmd": "uv run pytest tests/ -v -s -m '(account or basin or stream) and not access_tokens'"
+            "test_cmd": "uv run pytest tests/ -v -s -m '(account or basin or stream) and not access_tokens and not locations'"
           },
           {
             "name": "rust",
--- a/cli/src/apply.rs
+++ b/cli/src/apply.rs
@@ -122,8 +122,8 @@ fn stream_config_spec_to_sdk(config: StreamConfigSpec) -> s2_sdk::types::StreamC
     if let Some(timestamping) = config.timestamping {
         sdk_config = sdk_config.with_timestamping(timestamping_spec_to_sdk(timestamping));
     }
-    if let Some(delete_on_empty) = config.delete_on_empty {
-        sdk_config = sdk_config.with_delete_on_empty(delete_on_empty_spec_to_sdk(delete_on_empty));
+    if let Some(delete_on_empty) = config.delete_on_empty.and_then(delete_on_empty_spec_to_sdk) {
+        sdk_config = sdk_config.with_delete_on_empty(delete_on_empty);
     }
     sdk_config
 }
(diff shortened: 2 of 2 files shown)


quettabit:
@greptileai review pls

PR #498, opened 2026-06-02
fix(sdk): inconsistent ack deadline semantics in append session

### LINKED ISSUE #439: [Detail Bug] SDK AppendSession retries can trigger premature ack timeouts due to stale send timestamp
# Detail Bug Report

https://app.detail.dev/org_89d327b3-b883-4365-b6a3-46b6701342a9/bugs/bug_1a700f75-bd0a-49cd-b0cd-973e99112c07

Introduced in [#217](https://github.com/s2-streamstore/s2/pull/217) by @quettabit on Feb 16, 2026

# Summary
- **Context**: The `AppendSession` manages high-throughput appending with retry logic, where a stashed submission waits for channel capacity before being sent to the server.
- **Bug**: The `since` timestamp of a stashed submission is not reset when sending after a retry, causing premature acknowledgment timeouts.
- **Actual vs. expected**: After a retry, the timer deadline is calculated using the stale `since` value from before the retry, instead of the time when the submission was actually sent.
- **Impact**: Spurious `AckTimeout` errors that waste retry budget and can cause permanent session failure.

# Code with Bug

```rust
// In run_session() main loop (sdk/src/session/append.rs, lines 556-573)
input_tx_permit = input_tx.reserve(), if state.stashed_submission.is_some() => {
    let input_tx_permit = input_tx_permit
        .map_err(|_| AppendSessionError::ServerDisconnected)?;
    let submission = state.stashed_submission
        .take()
        .expect("stashed_submission should not be None");

    input_tx_permit.send(submission.input.clone());

    state.total_records += submission.input.records.len();
    state.inflight_bytes += submission.input_metered_bytes;

    timer.as_mut().fire_at(
        TimerEvent::AckDeadline,
        submission.since + ack_timeout,  // <-- BUG 🔴 uses pre-retry/stash timestamp, not wire-send time
        CoalesceMode::Earliest,
    );
    state.inflight_appends.push_back(submission.into());
}
```

# Explanation
A stashed submission can be created (capturing `since = Instant::now()`), then the session can hit a retryable ack-stream error while `input_tx.reserve()` is blocked waiting for channel capacity. On the subsequent retry, if `inflight_appends` is empty, `resend()` is skipped; when capacity becomes available, the stashed submission is sent and turned into an inflight append, but its `since` is not refreshed. The ack deadline is then scheduled from this stale time, making the timeout expire earlier than intended and potentially consuming retry attempts unnecessarily.

## Codebase Inconsistency
`resend()` explicitly resets `since` immediately before sending, implying `since` is intended to represent wire-send time; the stashed-submission send path does not.

```rust
// In resend() function (sdk/src/session/append.rs, lines 673-680)
if let Some(inflight_append) = state.inflight_appends.get_mut(resend_index) {
    inflight_append.since = Instant::now();  // Correctly reset before sending
    timer.as_mut().fire_at(
        TimerEvent::AckDeadline,
        inflight_append.since + ack_timeout,
        CoalesceMode::Earliest,
    );
    input_tx_permit.send(inflight_append.input.clone());
    resend_index += 1;
}
```

# Recommended Fix
Reset `since` when converting a stashed submission to an inflight append:

```rust
input_tx_permit = input_tx.reserve(), if state.stashed_submission.is_some() => {
    let input_tx_permit = input_tx_permit
        .map_err(|_| AppendSessionError::ServerDisconnected)?;
    let mut submission = state.stashed_submission
        .take()
        .expect("stashed_submission should not be None");

    input_tx_permit.send(submission.input.clone());

    state.total_records += submission.input.records.len();
    state.inflight_bytes += submission.input_metered_bytes;

    submission.since = Instant::now();
    timer.as_mut().fire_at(
        TimerEvent::AckDeadline,
        submission.since + ack_timeout,
        CoalesceMode::Earliest,
    );
    state.inflight_appends.push_back(submission.into());
}
```

# History
This bug was introduced in commit 3ae81ab. The commit moved the `s2-sdk` into the `s2` monorepo, bringing in 14,032 lines across 36 SDK files. The bug slipped in because the `resend()` function correctly resets `since = Instant::now()` when resending inflight appends (line 628), but the parallel code path for sending a stashed submission never applied the same reset—the developer handled one path correctly but omitted the timestamp reset on the other.

### LINKED ISSUE #441: [Detail Bug] Append session resend can raise premature AckTimeout due to non-extending timer deadline
# Detail Bug Report

https://app.detail.dev/org_89d327b3-b883-4365-b6a3-46b6701342a9/bugs/bug_dc6da1f3-9003-4697-bd4e-4ec7d28175d7

# Summary
- **Context**: The append session resend logic handles reconnection by resubmitting in-flight appends that haven't been acknowledged.
- **Bug**: The timer deadline during resend uses `Earliest` mode, which prevents the timer from extending when later appends are sent after earlier ones are acknowledged. Combined with stale `since` values, this causes premature `AckTimeout` errors.
- **Actual vs. expected**: When an earlier append is acknowledged during resend and a later append is then sent, the timer should extend to the later append's deadline. Instead, it stays at the earlier deadline.
- **Impact**: Timer fires based on a stale deadline; later appends timeout before their proper window.

# Code with Bug
```rust
// sdk/src/session/append.rs - resend() sets deadline using Earliest
if let Some(inflight_append) = state.inflight_appends.get_mut(resend_index) {
    inflight_append.since = Instant::now();  // Updated lazily, not upfront
    timer.as_mut().fire_at(
        TimerEvent::AckDeadline,
        inflight_append.since + ack_timeout,
        CoalesceMode::Earliest, // <-- BUG 🔴 prevents extending deadline for later resent appends
    );
    ...
}
```

```rust
// sdk/src/session/append.rs - process_ack() computes next deadline from front, which may still have stale `since`
if let Some(oldest_append) = state.inflight_appends.front() {
    timer.fire_at(
        TimerEvent::AckDeadline,
        oldest_append.since + ack_timeout,  // <-- BUG 🔴 can be stale during resend, interacting badly with coalescing
        CoalesceMode::Latest,
    );
}
```

# Explanation
During resend, `inflight_appends` initially contains appends with `since` timestamps from before the connection failure. Resend updates `since` only when each append is actually resent. If backpressure forces sequential sends (e.g., `max_unacked_batches = 1`), an ack for the first resent append can arrive before the next append is resent; `process_ack()` then looks at the front append (which may not have been resent yet) and computes a deadline from a stale `since`.

Separately, the resend loop arms the timer with `CoalesceMode::Earliest`. This makes the timer “sticky” to an earlier deadline: when a later append is finally resent with a later deadline, `Earliest` keeps the existing earlier deadline rather than extending it. This leads to premature `AckTimeout` for later resent appends.

Concrete trace (with `ack_timeout = 5s`, `max_unacked_batches = 1`): resend A at T=0 sets deadline to T=5. A is acked at T=2; resend B at T=2 should set deadline to T=7, but `Earliest` keeps T=5, so the timer fires at T=5 and returns `AckTimeout` even though B has only been in-flight for 3 seconds.

# Recommended Fix
- In resend loop, change `CoalesceMode::Earliest` to `CoalesceMode::Latest` so the deadline can extend when later appends are resent.
- At the start of resend, reset all in-flight appends’ `since` to `Instant::now()` so `process_ack()` doesn’t compute deadlines from pre-reconnect timestamps.

# History
Commit `8996947` fixed an `UnexpectedAck` panic in this same resend code path, indicating prior correctness issues in resend logic.

quettabit:
closes #439 

closes #441 

alters the previously mentioned expectation in https://github.com/s2-streamstore/s2/issues/358#issuecomment-4246260959

### DIFF
--- a/sdk/src/session/append.rs
+++ b/sdk/src/session/append.rs
@@ -560,6 +560,7 @@ async fn run_session(
                     .take()
                     .expect("stashed_submission should not be None");
 
+                let ack_deadline = Instant::now() + ack_timeout;
                 input_tx_permit.send(submission.input.clone());
 
                 state.total_records += submission.input.records.len();
@@ -567,10 +568,16 @@ async fn run_session(
 
                 timer.as_mut().fire_at(
                     TimerEvent::AckDeadline,
-                    submission.since + ack_timeout,
+                    ack_deadline,
                     CoalesceMode::Earliest,
                 );
-                state.inflight_appends.push_back(submission.into());
+                state.inflight_appends.push_back(InflightAppend {
+                    input: submission.input,
+                    input_metered_bytes: submission.input_metered_bytes,
+                    ack_tx: submission.ack_tx,
+                    ack_deadline,
+                    _permit: submission.permit,
+                });
             }
 
             cmd = state.cmd_rx.recv(), if state.stashed_submission.is_none() => {
@@ -587,7 +594,6 @@ async fn run_session(
                                 input_metered_bytes,
                                 ack_tx,
                                 permit,
-                                since: Instant::now(),
                             });
                         }
                     }
@@ -607,7 +613,6 @@ async fn run_session(
                             ack,
                             state,
                             timer.as_mut(),
-                            ack_timeout,
                         )?;
                     }
                     Some(Err(err)) => {
@@ -671,11 +676,11 @@ async fn resend(
                     .map_err(|_| AppendSessionError::ServerDisconnected)?;
 
                 if let Some(inflight_append) = state.inflight_appends.get_mut(resend_index) {
-                    inflight_append.since = Instant::now();
+                    inflight_append.ack_deadline = Instant::now() + ack_timeout;
                     timer.as_mut().fire_at(
                         TimerEvent::AckDeadline,
-                        inflight_append.since + ack_timeout,
-                        CoalesceMode::Earliest,
+                        inflight_append.ack_deadline,
+                        CoalesceMode::Latest,
                     );
                     input_tx_permit.send(inflight_append.input.clone());
                     resend_index += 1;
@@ -691,7 +696,6 @@ async fn resend(
                             ack,
                             state,
                             timer.as_mut(),
-                            ack_timeout,
                         )?;
                         resend_index = resend_index.checked_sub(1).ok_or_else(|| {
                             AppendSessionError::InvalidAck(
@@ -747,7 +751,6 @@ fn process_ack(
     ack: AppendAck,
     state: &mut SessionState,
     timer: Pin<&mut MuxTimer<N_TIMER_VARIANTS>>,
-    ack_timeout: Duration,
 ) -> Result<(), AppendSessionError> {
     let corresponding_append = state.inflight_appends.pop_front().ok_or_else(|| {
         AppendSessionError::InvalidAck(
@@ -787,7 +790,7 @@ fn process_ack(
     if let Some(oldest_append) = state.inflight_appends.front() {
         timer.fire_at(
             TimerEvent::AckDeadline,
-            oldest_append.since + ack_timeout,
+            oldest_append.ack_deadline,
             CoalesceMode::Latest,
         );
     } else {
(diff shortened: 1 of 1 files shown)


quettabit:
@greptileai review pls

PR #608, opened 2026-06-27
fix(cli): correlate benchmark events by run id to drop stale ones

### LINKED ISSUE #524: [Detail Bug] TUI Benchmark: Resetting during startup can cause stale benchmark events to complete/corrupt the next run
# Detail Bug Report

https://app.detail.dev/org_89d327b3-b883-4365-b6a3-46b6701342a9/bugs/bug_d83d2895-ab11-4cdf-8162-db281f182c3b

Introduced in [#120](https://github.com/s2-streamstore/s2/pull/120) by @infiniteregrets on Jan 29, 2026

# Summary
- **Context**: The benchmark feature in the TUI allows users to run performance tests on S2 streams, with progress displayed in real-time.
- **Bug**: A race condition allows a stale `BenchComplete` event from an abandoned benchmark to mark a new benchmark as complete.
- **Actual vs. expected**: When a user resets during the async stream creation window and starts a new benchmark, the old benchmark's completion event incorrectly terminates the new benchmark's UI state.
- **Impact**: The new benchmark's UI shows "Benchmark complete!" prematurely with wrong stats, while the benchmark continues running in the background. User makes decisions based on incorrect performance data.

# Code with Bug
```rust
KeyCode::Enter => {
    if state.config_field == BenchConfigField::Start {
        // ...
        state.config_phase = false;  // <-- BUG 🔴 Leaves a window where benchmark is not running yet
        self.start_benchmark(...);    // <-- async task starts; running still false until later event
    }
}
```

```rust
if !state.config_phase && !state.running {  // <-- BUG 🔴 True in the startup window after Enter
    match key.code {
        KeyCode::Char('r') => {
            let basin_name = state.basin_name.clone();
            self.screen = Screen::BenchView(BenchViewState::new(basin_name)); // New state, old task lives!
        }
        // ...
    }
}
```

```rust
Event::BenchComplete(result) => {
    if let Screen::BenchView(state) = &mut self.screen {
        state.running = false;  // <-- BUG 🔴 No validation/correlation to the currently active benchmark
        match result {
            Ok(stats) => {
                state.ack_latency = stats.ack_latency;
                state.e2e_latency = stats.e2e_latency;
                self.message = Some(StatusMessage {
                    text: "Benchmark complete!".to_string(),
                    level: MessageLevel::Success,
                });
            }
            Err(e) => {
                state.error = Some(e.to_string());
            }
        }
    }
}
```

```rust
fn start_benchmark(&mut self, ...) {
    // ...
    tokio::spawn(async move {
        // ... benchmark work ...
        let _ = tx.send(Event::BenchComplete(result)); // <-- BUG 🔴 Can arrive after user reset and new benchmark started
    });
    // JoinHandle is discarded - no way to abort old task
}
```

# Explanation
Pressing Enter flips the UI out of config mode immediately (`config_phase=false`) but the benchmark is not marked running until a later async event (stream creation completes). During this startup window, the `'r'` handler is enabled (`!config_phase && !running`) and it replaces the `BenchViewState` without canceling the spawned benchmark task.

Because benchmark events (including `BenchStreamCreated`, `BenchWriteSample`, `BenchReadSample`, `BenchPhaseComplete`, `BenchComplete`) carry no identifier tying them to the current benchmark run, any events still emitted by the abandoned task can be processed against the new `BenchViewState`. In particular, a stale `BenchComplete` sets `running=false` and overwrites latency stats/message, making the new run appear complete while it continues running in the background.

## Codebase Inconsistency
A stop/cancel mechanism exists (`bench_stop_signal`) and is used by the `'q'` handler to stop benchmarks before leaving the screen, but the `'r'` reset path does not stop/abort the existing task and therefore allows stale events to keep mutating the new benchmark state.

# Recommended Fix
Store the spawned benchmark task `JoinHandle` and abort it when resetting with `'r'` (and clear `bench_stop_signal`). This prevents *all* further events from an abandoned benchmark from being sent/handled. (Signaling stop via `bench_stop_signal` alone is insufficient because the task still sends `BenchComplete(result)` on exit.)

# History
This bug was introduced in commit 9fa3441. The TUI was added as a complete feature in a single commit (15,005 insertions across 12 files), and the 'r' key handler for resetting benchmark configuration was included from the start. The bug slipped in because the developer created a `bench_stop_signal` infrastructure for cancelling benchmark tasks (used by the 'q' key handler at line 7018), but forgot to use it in the 'r' handler. The code has remained unchanged since introduction - no subsequent commits have modified the 'r' handler or addressed the race condition.

infiniteregrets:
Closes #524.

## Problem

Pressing Enter to start a benchmark sets `config_phase = false` immediately, but `running` only becomes true later, when the async `BenchStreamCreated` event arrives. In that startup window the reset key (`r`) is enabled (`!config_phase && !running`) and replaced the `BenchViewState` **without stopping the spawned task**.

The abandoned task kept emitting events — most damagingly `BenchComplete` — which mutated the new run's state, marking it complete with stale stats while the real benchmark kept running in the background. The events carried no identifier tying them to a run, so there was no way to tell a stale event from a current one.

## Fix

Correlate benchmark events with the run they belong to:

- Tag every benchmark event with a `run_id` (`BenchRunId`). `App` holds the current `bench_run_id`; each new run bumps it.
- Every benchmark event handler drops events whose `run_id` is not the current run, so a stale event — including one already queued at the instant of reset — can't mutate the new state.
- On reset/leave, `stop_benchmark` signals the run to stop and detaches its task, leaving it alive to delete its temporary `_bench_*` stream (rather than aborting it mid-cleanup).
- Handlers also ignore events while in the config phase / not running.

Verified: build, `cargo clippy --all-features --all-targets -- -D warnings`, and `cargo test` all pass.

### DIFF
--- a/cli/src/tui/app.rs
+++ b/cli/src/tui/app.rs
@@ -19,7 +19,10 @@ use s2_sdk::types::{
 use tokio::sync::mpsc;
 
 use super::{
-    event::{BasinConfigInfo, BenchFinalStats, BenchPhase, BenchSample, Event, StreamConfigInfo},
+    event::{
+        BasinConfigInfo, BenchFinalStats, BenchPhase, BenchRunId, BenchSample, Event,
+        StreamConfigInfo,
+    },
     text_input::{
         TextInput, cursor_backspace, cursor_delete, cursor_insert, cursor_move_end,
         cursor_move_home, cursor_move_left, cursor_move_right,
@@ -1036,8 +1039,9 @@ pub struct App {
     should_quit: bool,
     /// Stop signal for the benchmark task
     bench_stop_signal: Option<Arc<AtomicBool>>,
-    /// Handle to the spawned benchmark task, so it can be aborted on reset
+    /// Handle to the current benchmark task
     bench_task: Option<tokio::task::JoinHandle<()>>,
+    bench_run_id: BenchRunId,
 }
 
 pub fn location_pill_idx(location: &str, custom_active: bool, names: &[&str]) -> usize {
@@ -1222,19 +1226,26 @@ impl App {
             should_quit: false,
             bench_stop_signal: None,
             bench_task: None,
+            bench_run_id: 0,
         }
     }
 
-    /// Hard-stop any in-flight benchmark task so it emits no further events.
-    /// Unlike the graceful stop on quit, the reset path reuses the BenchView
-    /// screen, so a late `BenchComplete` would corrupt the next run.
-    fn abort_benchmark(&mut self) {
+    fn next_benchmark_run_id(&mut self) -> BenchRunId {
+        self.bench_run_id = self.bench_run_id.wrapping_add(1);
+        self.bench_run_id
+    }
+
+    fn is_current_benchmark(&self, run_id: BenchRunId) -> bool {
+        run_id == self.bench_run_id
+    }
+
+    /// Ask any in-flight benchmark to stop and detach its task.
+    /// The task is left alive so it can delete the temporary stream.
+    fn stop_benchmark(&mut self) {
         if let Some(stop) = self.bench_stop_signal.take() {
             stop.store(true, Ordering::Relaxed);
         }
-        if let Some(task) = self.bench_task.take() {
-            task.abort();
-        }
+        self.bench_task.take();
     }
 
     /// Create an S2 client from the given access token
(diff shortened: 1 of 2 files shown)


infiniteregrets:
@greptile-apps review

PR #641, opened 2026-07-15
fix(lite): make provision_stream exists-outcomes durably visible

infiniteregrets:
## Problem

Concurrent appends (or reads) that auto-create the same fresh stream — basin has `create_stream_on_append` / `create_stream_on_read` — race in `stream_handle_with_auto_create`. Exactly one racer creates the stream cleanly; the others split into two fates:

- **Conflict losers**: their provision txn read the stream meta as absent, the winner committed first → SSI read-write conflict → `TransactionConflict` → HTTP 409, which clients retry and succeed. Fine.
- **Already-exists losers**: their txn began after the winner's memtable write → `provision_stream`'s existence check (`db_txn_get`, `DurabilityLevel::Memory`) sees the meta → `StreamAlreadyExists`, correctly swallowed. But the immediately following `start_streamer` reads that same meta at `DurabilityLevel::Remote`, whose watermark can lag the winner's `await_durable` commit by up to one flush interval → `StreamNotFound` → **HTTP 404, non-retryable**. For appends, clients that treat 404 as terminal silently lose the write.

The server proves the stream exists and denies it exists within the same request. Repro: 4 concurrent unary appends to a fresh auto-create stream → 1 OK, others 404 (some 409→retry→OK). Immediate retry of the 404s succeeds; once the stream exists the race is gone. S2 cloud is unaffected (verified 30/30 concurrent auto-create appends).

```
timeline ─────────────────────────────────────────────────────▶
winner:   get(meta)=None ── put ── commit(await_durable) ══flush══ done ── refetch OK
loser A:  get(meta)=None ─────── commit → CONFLICT → 409 → client retries → OK
loser B:            get(meta)=Some (Memory) → AlreadyExists → refetch (Remote) → 404 ← bug
```

## Verification

- New regression test `concurrent_appends_auto_create_stream_without_spurious_not_found` reproduces the 404 in-process on unfixed code (fails on first round) and passes with the fix.
- Full `s2-lite` suite: 120/120; `cargo fmt` + `clippy --lib --tests` clean.
- End-to-end against the rebuilt server binary: TypeScript SDK repro went 30/30 concurrent auto-create appends with zero 404s (conflict losers succeed via SDK retry), matching observed cloud behavior.

### DIFF
--- a/lite/src/backend/core.rs
+++ b/lite/src/backend/core.rs
@@ -89,7 +89,6 @@ impl Backend {
         let reason = match rx.await {
             Ok(Ok(_)) => return Ok(()),
             Ok(Err(reason)) => reason,
-            // Callback dropped uninvoked: abrupt runtime shutdown.
             Err(_) => slatedb::CloseReason::Clean,
         };
         Err(slatedb::Error::closed(
@@ -376,8 +375,6 @@ impl Backend {
                             }
                         }
                     }
-                    // provision_stream guarantees the outcome is durably
-                    // visible, so start_streamer's Remote-level read sees it.
                     let client = self.streamer_client_guarded(basin, stream).await?;
                     let encryption = resolve_encryption(client.cipher())?;
                     Ok(StreamHandle {
--- a/lite/src/backend/core.rs
+++ b/lite/src/backend/core.rs
@@ -1,4 +1,4 @@
-use std::sync::Arc;
+use std::{sync::Arc, time::Duration};
 
 use bytesize::ByteSize;
 use dashmap::DashMap;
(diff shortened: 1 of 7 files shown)


infiniteregrets:
@greptile-apps review

lite/src/backend/core.rs
```diff
@@ -557,4 +576,66 @@ mod tests {
             _ => panic!("expected initializing slot to remain unchanged"),
         }
     }
+
+    #[tokio::test(flavor = "multi_thread")]
+    async fn concurrent_appends_auto_create_stream_without_spurious_not_found() {
```
shikhar, line 581: does this fail without the fix?

shikhar (approved):

lite/src/backend/core.rs
infiniteregrets:
```
running 1 test
  test backend::core::tests::concurrent_appends_auto_create_stream_without_spurious_not_found ...
  FAILED

  failures:

  ---- backend::core::tests::concurrent_appends_auto_create_stream_without_spurious_not_found stdout
  ----

  thread 'backend::core::tests::concurrent_appends_auto_create_stream_without_spurious_not_found'
  (127981022) panicked at lite/src/backend/core.rs:639:31:
  concurrent auto-create append failed: StreamNotFound(StreamNotFoundError { basin: autocreate,
  stream: fresh-0 })
  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


  failures:
      backend::core::tests::concurrent_appends_auto_create_stream_without_spurious_not_found

  test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 119 filtered out; finished in 0.32s

  error: test failed, to rerun pass `-p s2-lite --lib`
  ```

infiniteregrets: it does

PR #643, opened 2026-07-16
feat(lite): emit tail in read session SSE pings

infiniteregrets:


api/src/v1/stream/sse.rs
```diff
@@ -182,4 +185,6 @@ impl AsRef<str> for DoneEventData {
 #[cfg_attr(feature = "utoipa", derive(utoipa::ToSchema))]
 pub struct PingEventData {
     pub timestamp: u64,
```
shikhar, line 188: May as well doc this, `Server time when the ping was sent`

shikhar (approved):

PR #650, opened 2026-07-19
feat(sdk)!: expose caught-up-to-tail signal on read sessions

infiniteregrets:
## Summary

- expose caught-up state and an awaitable tail on Rust read sessions
- update the signal when the matching batch is delivered, including heartbeats, filtered commands, and retries
- preserve the CLI stream interface and add focused unit and live correctness coverage

## Validation

- `cargo +stable nextest run --workspace --all-features --exclude s2-testcontainers -E 'not ((package(s2-cli) & binary(integration)) or (package(s2-sdk) & (binary(account_ops) or binary(basin_ops) or binary(metrics_ops) or binary(stream_ops))))'` (673 passed)
- `cargo +stable clippy --workspace --all-features --all-targets --exclude s2-testcontainers -- -D warnings --allow deprecated`
- `RUSTDOCFLAGS='-D warnings' cargo +stable doc -p s2-sdk --all-features --no-deps`
- `just fmt`

### DIFF
--- a/sdk/src/session/read.rs
+++ b/sdk/src/session/read.rs
@@ -230,41 +230,11 @@ impl ReadSession {
 
     /// Return a future for the current or next caught-up tail.
     ///
-    /// It is ready immediately when already caught up and remains pending across retries.
-    /// Once resolved, it keeps that tail if the session later falls behind.
-    /// Keep reading batches while waiting. Call again after falling behind. Returns
-    /// [`CaughtUpError`] if the read fails or ends first.
-    ///
-    /// # Example
-    ///
-    /// Drive the session stream while waiting for the caught-up signal:
-    ///
-    /// ```rust
-    /// use futures_util::StreamExt;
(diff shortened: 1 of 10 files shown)


sdk/src/session/read.rs
```diff
@@ -133,17 +363,13 @@ pub async fn read_session(
...
-                    }
+                    yield Ok(ReadUpdate::from_batch(batch, ignore_command_records));
                 }
                 Some(Err(err)) => {
                     batches = None;
-                    if can_retry(&err, &mut retry_backoff).await {
+                    if let Some(backoff) = retry_delay(&err, &mut retry_backoff) {
+                        yield Ok(ReadUpdate::behind());
+                        tokio::time::sleep(backoff).await;
                         continue;
```
infiniteregrets, line 388: when retrying we mark caughtup state to be behind, lmk if we wanna change this

shikhar: no it makes sense I think

PR #651, opened 2026-07-20
build: use Minimus Rust builder for Docker image

emiioan:
Changes:
- replace rust:latest with public reg.mini.dev/rust:latest-dev
- retain root builder semantics for BuildKit cache compatibility
- keep the final Distroless runtime non-root with no ECR dependency

Tests:
- docker build --target runtime --tag s2:minimus-test .
- S2 Lite health, basin and stream creation, three-record append/read
- just fmt
- RUSTUP_TOOLCHAIN=stable just test (666 passed, 1 skipped)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,7 @@
 # Build stage
-FROM rust:latest AS builder
+FROM reg.mini.dev/rust:latest AS builder
+
+USER root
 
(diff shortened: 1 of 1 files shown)


cristi- (approved):
lg!

emiioan:
@greptileai

PR #660, opened 2026-07-23
refactor(cli)!: remove interactive TUI

infiniteregrets:
## Summary

- remove the interactive TUI and its `-i`/`--interactive` entry point
- remove the `ratatui` and `crossterm` dependency graph
- remove the now-unused direct `chrono` dependency
- retain benchmark terminal-width detection with the lightweight `terminal_size` crate
- remove the latency helper that was only used by the TUI

## Why

The interactive terminal UI is no longer part of the supported CLI surface. Removing it simplifies the CLI, eliminates roughly 16,000 lines of UI-specific code, and reduces the dependency graph.

## Impact

This is a breaking CLI change: `s2 -i` and `s2 --interactive` are no longer accepted. All regular subcommands remain available.

## Validation

- `just fmt`
- `cargo +stable nextest run --workspace --all-features --exclude s2-testcontainers -E 'not ((package(s2-cli) & binary(integration)) or (package(s2-sdk) & (binary(account_ops) or binary(basin_ops) or binary(metrics_ops) or binary(stream_ops))))'` — 684 passed, 1 skipped
- `cargo metadata --locked --format-version 1 --no-deps`
- verified `ratatui` and `crossterm` are absent from the `s2-cli` dependency tree
- `cargo-machete` — no unused `s2-cli` dependencies
- `cargo udeps --locked -p s2-cli --bin s2 --test cli --all-features` — no unused normal dependencies; reported only target-exclusion false positives for dev dependencies used by unit or integration tests

`just test` and the all-targets udeps run under the repository's current nightly toolchain hit an existing rustc ICE while compiling the excluded CLI integration target; the equivalent test selection passes under stable.


shikhar (approved):
suggest running machete / udeps to double-check on unused deps

PR #682, opened 2026-08-04
feat(sdk): support refreshable access token providers

infiniteregrets:
## Summary

- add an internal, feature-gated access token provider to the SDK
- resolve and mark bearer credentials as sensitive for each request attempt
- invalidate the exact credential rejected by an S2 `401/authn` response and retry within the existing retry budget
- reconnect read and append sessions with refreshed credentials while preserving their existing resume and append retry policies
- preserve the existing static access-token API and behavior

## Why

The SDK currently installs one bearer token when a client is constructed. Browser OAuth access tokens expire, but request retries and streaming reconnects are owned by the SDK. This change introduces a small internal callback so the CLI can supply a current opaque bearer token without teaching the SDK about OAuth, Clerk, JWTs, expiration, or refresh tokens.

The provider API is behind the existing `_hidden` feature and marked `#[doc(hidden)]`; it is not part of the supported public SDK surface.

## Safety

- refresh is triggered only for `401` responses whose S2 error code is exactly `authn`
- invalidation includes the exact rejected token, avoiding stale responses invalidating newer credentials
- provider retries remain bounded by the existing SDK retry configuration
- append retries continue to respect `AppendRetryPolicy`
- authorization headers and credential debug output are redacted
- transient provider failures are retryable; permanent failures return immediately

## Validation

- `just fmt`
- `cargo clippy --locked -p s2-sdk --lib --features _hidden -- -D warnings --allow deprecated`
- `cargo nextest run --locked -p s2-sdk --all-features -E 'not (binary(account_ops) or binary(basin_ops) or binary(metrics_ops) or binary(stream_ops))'` — 109 passed
- `cargo test --locked -p s2-sdk --lib` — 106 passed on the normal static-token build
- `cargo check --locked -p s2-cli`

The repository-wide `just clippy` / `just test -p s2-sdk` wrappers currently hit an unrelated nightly compiler ICE while compiling `cli/tests/integration.rs:51`. Live SDK integration tests were not run because they require `S2_ACCESS_TOKEN`.

### DIFF
--- a/sdk/src/session/read.rs
+++ b/sdk/src/session/read.rs
@@ -22,7 +22,7 @@ use crate::{
     error::{ReadError, RequestError},
     retry::RetryBackoff,
     types::{
-        EncryptionKey, MeteredBytes, ReadBatch, ReadInput, ReadSessionConfig,
+        AccessTokenMode, EncryptionKey, MeteredBytes, ReadBatch, ReadInput, ReadSessionConfig,
         ReadSessionRetryPolicy, StreamName, StreamPosition,
     },
 };
@@ -365,7 +365,7 @@ pub async fn read_session(
     let mut end: ReadEnd = stop.into();
     let retry_policy = config.retry_policy;
     let mut retry_backoff = retry_builder(&client.config.retry).build();
-    let refreshable_auth = client.config.access_token.is_dynamic();
+    let access_token_mode = client.config.access_token.mode();
     let baseline_wait = end.wait;
     let mut last_tail_at: Option<Instant> = None;
     let initial_resume_seq_num = if start.clamp == Some(true) {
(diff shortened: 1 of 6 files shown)


sdk/src/session/read.rs
```diff
@@ -557,23 +576,54 @@ fn retry_delay(
...
-            is_retryable = err.is_retryable(),
+            is_retryable,
             retries_exhausted = backoffs.is_exhausted(),
             "not retrying read session"
         );
         None
     }
 }
 
+fn is_retryable_with_auth_refresh(error: &ReadSessionFailure, refreshable_auth: bool) -> bool {
```
shikhar: just 1 call-site, factored out with a bool param for sake of trivialish unit test, just inline (also in read session)

sdk/src/session/read.rs
```diff
@@ -530,8 +547,10 @@ fn retry_delay(
     err: &ReadSessionFailure,
     backoffs: &mut RetryBackoff,
     retry_policy: ReadSessionRetryPolicy,
+    refreshable_auth: bool,
```
shikhar: don't like bool args, enum might have been overkill but i see this thing is needed in 3 modules, so probably makes sense

shikhar (approved):

PR #700, opened 2026-08-13
feat(api): s2s reconnect advice flag and server_draining error code

infiniteregrets:
## Summary

- Add a reconnect-advised bit (`0x10`) to the S2S frame flag byte. A server that is about to terminate sets it on regular response frames so clients can proactively reconnect elsewhere. The protobuf body is unchanged, so clients that ignore the bit keep working.
- `s2s::advise_reconnect` stamps the bit onto an already encoded frame without copying, and leaves terminal frames unchanged.
- `CompressedData::reconnect_advised` surfaces the bit on decode, so the SDK can observe the advice through its existing `FrameDecoder` path.
- Add `ErrorCode::ServerDraining` (`server_draining`, 503) as the terminal error for sessions ended by a draining server. It is sent only after all accepted inputs have been acknowledged, so it is retryable and guarantees no side effects for unacknowledged inputs.

### DIFF
--- a/api/src/v1/error.rs
+++ b/api/src/v1/error.rs
@@ -39,6 +39,7 @@ pub enum ErrorCode {
     RateLimited,
     RequestTimeout,
     ResourceAlreadyExists,
+    ServerDraining,
     Storage,
     StreamDeletionPending,
     StreamNotFound,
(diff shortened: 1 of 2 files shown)


shikhar (approved):
pls double check none of the s2s clients are trying to interpret this bit e.g. asserting 0

PR #703, opened 2026-08-17
feat(sdk): act on s2s reconnect advice in append and read sessions

infiniteregrets:


sdk/src/session/append.rs
```diff
@@ -591,8 +626,16 @@ async fn run_session(
 
     let timer = MuxTimer::<N_TIMER_VARIANTS>::default();
     tokio::pin!(timer);
+    let mut advised = false;
```
sgbalogh: `reconnect_advised` would be clearer

sdk/src/session/append.rs
```diff
@@ -591,8 +626,16 @@ async fn run_session(
...
 
     loop {
+        // Stop feeding a draining server. A close already in progress is
+        // cheaper to finish here than to move onto a fresh connection.
+        if reconnect.is_advised() && state.close_tx.is_none() {
+            advised = true;
```
sgbalogh: instead of having the 
```rust
if advised {
 // ...
}
```
bit after the loop, could we hoist it into a fn -- similar to the `resend` fn -- and instead:
```rust
  if reconnect.is_advised() && state.close_tx.is_none() {
      drain_for_reconnect(
          input_tx,
          acks,
          state,
          timer.as_mut(),
          ack_timeout,
      )
      .await?;

      return Ok(SessionOutcome::ReconnectAdvised);
  }

```

with that fn being something like
```rust
  async fn drain_for_reconnect(
      input_tx: mpsc::Sender<AppendInput>,
      mut acks: Streaming<AppendAck>,
      state: &mut SessionState,
      mut timer: Pin<&mut MuxTimer<N_TIMER_VARIANTS>>,
      ack_timeout: Duration,
  ) -> Result<(), AppendSessionError>
```

sdk/src/session/read.rs
```diff
@@ -513,15 +549,20 @@ async fn session_inner(
...
-) -> Result<InternalStreaming<ReadBatch>, ReadSessionFailure> {
+    reconnect: ReconnectAdvice,
+) -> Result<InternalStreaming<ReadItem>, ReadSessionFailure> {
     let mut batches = client
-        .read_session(&name, start, end, encryption.as_ref())
+        .read_session(&name, start, end, encryption.as_ref(), reconnect.clone())
         .await?;
     Ok(Box::pin(try_stream! {
         loop {
             match timeout(Duration::from_secs(20), batches.next()).await {
```
sgbalogh: not a part of this PR I know, but lets move this duration into a const with a comment, so its less magic-y and clearly related to the document s2s heartbeat max https://s2.dev/docs/api/protocol#data-flow (plus some buffer)

sdk/src/session/read.rs
```diff
@@ -439,7 +454,28 @@ pub async fn read_session(
                 .next()
                 .await
             {
-                Some(Ok(batch)) => {
+                Some(Ok(ReadItem::ReconnectAdvised)) => {
+                    batches = None;
+                    if end.count == Some(0) || end.bytes == Some(0) {
```
sgbalogh: this is confusing -- maybe lets have a util fn that clarifies that we are checking if the read end condition is already satisfied, like `read_limits_exhausted(&end)` or something, and maybe a comment stating that this is an optimization, to avoid a useless reconnect for a read that was already satisfied at the moment we received the reconnect signal

sdk/src/session/append.rs
infiniteregrets: not needed, since https://github.com/s2-streamstore/s2/pull/703#discussion_r3807640886

sdk/src/session/read.rs
infiniteregrets: PTAL

sdk/src/session/read.rs
infiniteregrets: thanks, please check!

sdk/src/client.rs
```diff
@@ -411,6 +411,12 @@ impl StreamingResponse {
 pub trait RequestExecutor: Send + Sync {
     async fn execute_unary(&self, request: Request) -> Result<UnaryResponse, HttpError>;
     async fn init_streaming(&self, request: Request) -> Result<StreamingResponse, HttpError>;
+
+    /// Discard pooled connections to `host` so the next request opens a fresh
+    /// one. Requests already in flight keep their connection.
+    async fn rotate(&self, host: &str) {
```
sgbalogh: why is this default needed?

sdk/src/client.rs
```diff
@@ -906,6 +912,16 @@ where
         let (client, permit) = self.checkout(request.authority()).await;
         init_streaming_with(&client, request, permit).await
     }
+
+    async fn rotate(&self, host: &str) {
```
sgbalogh: this will clear all pool clients for the host when called, even if they are on "good" hosts that are not going down and would not be sending reconnect advice right?

maybe we should only rotate the connection that received the advice, which we could track by maintaining some connection id? so rotation only happens for a single connection at a time, if at least one session on that connection received reconn advice. my thinking is that this could reduce unnecessary churn (costly and wasteful to have perfectly good connections re-establish, even if it wouldn't sever them during a session; plus maybe theres a potential for thundering herd here)

on that last point, lets also make sure to test out in DST that `rotate` on a connection is graceful for concurrent sessions that are on the same connection

sdk/src/session/append.rs
```diff
@@ -593,6 +626,11 @@ async fn run_session(
     tokio::pin!(timer);
 
     loop {
+        if reconnect.is_advised() && state.close_tx.is_none() {
```
sgbalogh: from my codex, seems legit?
```
state.close_tx.is_none() prevents acting on advice during close. Once acknowledgements finish, the outcome is Closed, so transport rotation
    is skipped. Pool invalidation should be independent of whether the append session itself needs another connection.
```

sdk/src/session/read.rs
```diff
@@ -439,7 +461,36 @@ pub async fn read_session(
                 .next()
                 .await
             {
-                Some(Ok(batch)) => {
+                Some(Ok(ReadItem::ReconnectAdvised)) => {
+                    batches = None;
+                    // Avoid a useless reconnect for a read that was already
+                    // satisfied when the advice arrived.
+                    if read_limits_exhausted(&end) {
+                        break;
```
sgbalogh, line 488: should also rotate transport in this case I think

sdk/src/client.rs
infiniteregrets: clients that dont pool can have this noop, but i guess it would just make sense to defer it to the client? its just stylistic, ill rm it

sdk/src/client.rs
infiniteregrets: exercising this in DST now, will lyk thanks!

infiniteregrets: okay this issue of thundering herd didnt come to me as i was relying on multiplexing where the number of hosts in the pool is minimal... but it makes sense to have a mechanism to prevent this

sdk/src/client.rs
```diff
@@ -741,6 +772,7 @@ struct PooledClient<C> {
 impl<C> PooledClient<C> {
     fn new(client: HyperClient<C, BoxBody>) -> Self {
         Self {
+            id: ConnectionId::next(),
             client: Arc::new(client),
             active_requests: Arc::new(AtomicUsize::new(0)),
             idle_since: Arc::new(Mutex::new(Some(Instant::now()))),
```
infiniteregrets: every pooled conn now gets an ID

sdk/src/client.rs
infiniteregrets: i think having a poison handle on the session itself would be a better design

sdk/src/client.rs
```diff
@@ -793,34 +852,47 @@ where
...
 
-    async fn reap_idle_clients(&self) {
+    /// Drop the pooled client identified by `id` so no new request reuses it.
+    /// Clients pinned to servers that are not going away stay pooled, requests
+    /// already in flight keep their connection, and poisoning the same
+    /// connection again is a no-op.
+    fn poison(&self, host: &str, id: ConnectionId) {
+        let mut clients = self.clients.write().unwrap();
+        let pooled = clients.len();
+        clients.retain(|pooled| pooled.id != id);
```
infiniteregrets, line 892: we could switch to a hashmap, but as i have understood a keyed container would be O(1) removal  but checkout on the per-request hot path scans for a free permit regardless and loses the vector "locality"...

sdk/src/client.rs
```diff
@@ -903,13 +978,23 @@ where
...
 
     async fn init_streaming(&self, request: Request) -> Result<StreamingResponse, HttpError> {
-        let (client, permit) = self.checkout(request.authority()).await;
-        init_streaming_with(&client, request, permit).await
+        let host = request.authority().to_owned();
+        let pool = self.get_or_create_host_pool(&host).await;
+        let (client, permit, id) = pool.checkout();
+        // Weak: the handle can outlive the pool (a session holds it for the
+        // connection's lifetime) and must not keep a reaped pool alive.
+        let weak = Arc::downgrade(&pool);
```
infiniteregrets, line 996: ok so claude suggests we keep this a weak reference, which is pretty cool idea given this:

>  t0  you open a read session        → response carries handle (Arc vote #2, map holds vote #1)
t1  you drop the whole S2 client   → map dropped, vote #1 gone
                                      everything about the pool SHOULD be freed now
t2  ...but your session/stream object is still alive somewhere,
    holding a handle that will almost certainly never fire
    → its Arc vote keeps the dead HostPool (Vec, connector) in memory
t3  forever, or until the session ends

sdk/src/client.rs
sgbalogh: > poison handle on the session 

called by session that receives the reconnect advice, but would taint the connection as a whole right?

infiniteregrets: yeah,,

```
pub(crate) fn advise(&self) {
        // The advice bit repeats on every frame while the server drains; only
        // the first one poisons the pooled connection.
        if !self.0.advised.swap(true, Ordering::AcqRel)
            && let Some(poison) = self.0.poison.get()
        {
            poison.poison();
        }
    }
  ```
  
  this bool is set and the conn removed, and all existing ones just drain and reconnected

Comment (infiniteregrets) [reply sdk/src/client.rs]: i just added this change, but its using a sync lock...

Comment (infiniteregrets) [reply sdk/src/client.rs]: looks like we are all good with using a sync lock here as we dont hold the lock across a bounded task , was comparing this with hyper-util and how they do the pool, they seem to work with sync locks as well https://github.com/hyperium/hyper-util/blob/v0.1.20/src/client/pool/cache.rs#L55

File: sdk/src/session/append.rs (hunk 7 above)
Comment (infiniteregrets) [reply sdk/src/session/append.rs]: pool invalidation logic was decooupled from this stuff

File: sdk/src/session/read.rs (hunk 8 above)
Comment (infiniteregrets) [reply sdk/src/session/read.rs:488]: this is fixed with the rewrite of how we handle the rotation. btw i renamed it to poison instead of rotate, seems more consistent with rust

https://github.com/smithy-lang/smithy-rs/blob/main/rust-runtime/aws-smithy-runtime-api/src/client/connection.rs#L23

**Comment by infiniteregrets** [comment]
ran against a sandbox cell, observed p99:


  | | p99 values | median | worst |
  |---|---|---|---|
  | OLD (bit ignored, baseline SDK) | 521, 470, 495, 542, 637, 510 | 516ms | 637ms |
  | NEW (bit acted on, #703) | 372, 318, 325, 366, 314, 930 | **346ms** | 930ms |

did kubectl rollouts at certain time points

File: sdk/src/reconnect.rs [hunk 12]
```diff
@@ -0,0 +1,59 @@
...
+        if !self.0.advised.swap(true, Ordering::AcqRel)
+            && let Some(poison) = self.0.poison.get()
+        {
+            poison.poison();
+        }
+    }
+
+    /// Capture the poison handle for the pooled connection serving this
+    /// session, once known.
+    pub(crate) fn capture_poison_handle(&self, poison: Option<PoisonHandle>) {
```
sgbalogh: maybe `set_connection_poison_handle`?

also I think neater if it accepts a `handle: PosionHandle` instead of `Option<PoisonHandle>`, and just use an `if let Some` at the callers

edit: see https://github.com/s2-streamstore/s2/pull/703#discussion_r3869468511 -- we can avoid this fn entirely

sdk/src/api.rs
```diff
@@ -497,6 +500,9 @@ impl BasinClient {
                 loop {
                     match decoder.decode(&mut buffer) {
                         Ok(Some(SessionMessage::Regular(msg))) => {
+                            if msg.reconnect_advised() {
+                                reconnect.advise();
```
sgbalogh, line 507: lets just cache status locally and call this once per response attempt? since `advise()` involves an atomic op

```rust
let mut reconnect_advice_seen: bool = false;

if !reconnect_advice_seen && msg.reconnect_advised() {
    reconnect_advice_seen = true;
    reconnect.advise();
}
```
and similar for read sesh

sdk/src/client.rs
```diff
@@ -369,27 +369,76 @@ impl UnaryResponse {
...
         }
     }
 
     pub fn status(&self) -> StatusCode {
         self.status
     }
 
+    /// Take the handle for poisoning the pooled connection this response is
+    /// being served on, when the executor pools connections.
+    pub fn take_poison_handle(&mut self) -> Option<PoisonHandle> {
```
sgbalogh: I think this can also return `PoisonHandle`, and that `.take().expect("only taken once")` is warranted, as it would be an invariant violation for this fn to ever be called >1

sgbalogh: played around with this... seems cleaner to not do this injection stuff with `take_poison_handle` and https://github.com/s2-streamstore/s2/pull/703#discussion_r3869265951

instead, can have `StreamingResponse` transform into a tuple of the bytestream and the handle, wdyt

```rust
 pub struct StreamingResponse {
      status: StatusCode,
      headers: HeaderMap,
      body: Incoming,
      permit: RequestPermit,
      poison_handle: PoisonHandle,
  }

  impl StreamingResponse {
      pub(crate) fn into_stream(
          self,
      ) -> (
          impl Stream<Item = Result<Bytes, HttpError>>,
          PoisonHandle,
      ) {
          let Self {
              body,
              permit,
              poison_handle,
              ..
          } = self;

          let stream =
              http_body_util::BodyStream::new(body).filter_map(move |result| {
                  let _ = &permit;
                  std::future::ready(match result {
                      Ok(frame) => frame.into_data().ok().map(Ok),
                      Err(error) => Some(Err(HttpError::Receive(error))),
                  })
              });

          (stream, poison_handle)
      }
  }

// ... in api.rs:

  let response = match response.into_result().await {
      Ok(response) => response,
      Err(error) => {
          // ...
          return Err(error);
      }
  };

  let (mut bytes_stream, poison_handle) = response.into_stream();
  reconnect.capture_poison_handle(poison_handle);
```

sdk/src/client.rs
sgbalogh: yeah this is a good solution!

sdk/src/client.rs
infiniteregrets: hmm yeah this looks much cleaner

sdk/src/reconnect.rs
```diff
@@ -12,7 +12,7 @@ use tokio::time::Instant;
 pub(crate) const MAX_ADVISED_RECONNECTS: usize = 1;
 
 /// Gap after which the attempt count resets.
-pub(crate) const ADVISED_RECONNECT_IDLE: Duration = Duration::from_secs(5);
+pub(crate) const ADVISED_RECONNECT_IDLE: Duration = Duration::from_secs(10);
```
sgbalogh: I wonder if this should be even higher, like 60 seconds? if the initial reconnect does not work, it is much more likely to be due to some odd pathological issue, and a longer reconnect period would protect against degraded QoS (from reconnect churn) during that period

sgbalogh: (i.e. should be high enough not to cause needless stress, low enough to not span a legitimate normal succession of frontend deployment rollouts)

infiniteregrets: this reminds me, on receiving a terminal, the reconnect shouldnt consume the ordinary retry budget right within this time frame of 60 seconds?

sgbalogh (approved):

infiniteregrets:
@greptile-apps review

PR #705, opened 2026-08-18
fix(cli): reject symlinked credential directory on the write path

### LINKED ISSUE #701: [Detail Bug] CLI: File-based credential storage writes through symlinked config directory, leaking tokens
# Detail Bug Report

https://app.detail.dev/org_89d327b3-b883-4365-b6a3-46b6701342a9/bugs/bug_2fdd386d-06dd-417c-8ee5-ac7e2c1ca686

Introduced in [#683](https://github.com/s2-streamstore/s2/pull/683) by @infiniteregrets on Aug 5, 2026

# Summary
- **Context**: The credential store manages storage of OAuth tokens and access tokens for the S2 CLI. Users can choose file-based storage with `--insecure-storage` flag instead of the default OS keyring.
- **Bug**: The write path (`write_private_file`) follows symlinks without validation, while the read path (`secure_private_file_for_read`) rejects symlinks. An attacker with write access to `~/.config/` can create a symlink `~/.config/s2` → `/attacker/location`, causing credentials to be silently written to an attacker-controlled location.
- **Actual vs. expected**: File-based credentials should be written to `~/.config/s2/` with symlink validation matching the read path. Instead, they are written through symlinks to arbitrary locations.
- **Impact**: **Credential leakage** - OAuth tokens and access tokens written to attacker-controlled locations, enabling account takeover for users who opt into file-based storage.

# Code with Bug
```rust
// cli/src/credential_store.rs
fn write_private_file(path: &Path, bytes: &[u8]) -> Result<(), CredentialStoreError> {
    let parent = path.parent().ok_or(CredentialStoreError::InvalidPath)?;
    fs::create_dir_all(parent).map_err(|source| CredentialStoreError::CredentialFile {  // <-- BUG 🔴 follows symlinked parent directory
        action: "create the parent directory for",
        source,
    })?;
    secure_directory(parent)?;  // <-- BUG 🔴 uses permission-setting APIs that follow symlinks

    let mut temp = tempfile::NamedTempFile::new_in(parent).map_err(|source| {  // <-- BUG 🔴 temp file created in symlink target
        CredentialStoreError::CredentialFile {
            action: "create",
            source,
        }
    })?;
    // ...
    temp.persist(path)  // <-- BUG 🔴 persists via symlinked path
        .map_err(|error| CredentialStoreError::CredentialFile {
            action: "replace",
            source: error.error,
        })?;
    Ok(())
}
```

# Explanation
- `write_private_file()` operates on `parent` using `create_dir_all()`, `secure_directory()`, and `NamedTempFile::new_in(parent)`; these operations follow symlinks.
- If `~/.config/s2` is a symlink to an attacker-controlled directory, the write path will create/persist the credential file in the attacker-controlled directory.
- The read path does symlink-safe metadata checks (`symlink_metadata`) and rejects symlink parents, so the user may see a read error later—after the credential was already written and exfiltrated.

## Codebase Inconsistency
The read path explicitly avoids symlink following and rejects non-directories:
```rust
// cli/src/credential_store.rs
#[cfg(unix)]
fn secure_private_file_for_read(path: &Path) -> Result<(), CredentialStoreError> {
    let parent = path.parent().ok_or(CredentialStoreError::InvalidPath)?;
    let directory = fs::symlink_metadata(parent).map_err(|source| {
        // ...
    })?;
    if !directory.file_type().is_dir() {
        return Err(CredentialStoreError::UnsafeCredentialFile(
            "the parent path is not a directory",
        ));
    }
    // ...
}
```

# Failing Test
A repro test was added and currently passes, proving the write follows a symlinked parent directory:
```rust
#[test]
fn write_private_file_follows_symlink_parent_directory() {
    let directory = tempfile::tempdir().unwrap();
    let attacker_controlled = directory.path().join("attacker");
    let symlink_to_attacker = directory.path().join("s2");

    fs::create_dir(&attacker_controlled).unwrap();
    fs::set_permissions(&attacker_controlled, fs::Permissions::from_mode(0o777)).unwrap();

    symlink(&attacker_controlled, &symlink_to_attacker).unwrap();

    let credential_path = symlink_to_attacker.join("credential.json");

    let result = write_private_file(&credential_path, b"secret credentials");

    assert!(result.is_ok(), "write_private_file should succeed");

    let attacker_file = attacker_controlled.join("credential.json");
    assert!(
        attacker_file.exists(),
        "Credential was NOT written to attacker-controlled directory"
    );

    let content = fs::read(&attacker_file).unwrap();
    assert_eq!(content, b"secret credentials", "Attacker has the credentials");

    let symlink_meta = fs::symlink_metadata(&symlink_to_attacker).unwrap();
    assert!(
        symlink_meta.file_type().is_symlink(),
        "Symlink still exists (not replaced)"
    );
}
```
Test output:
```
running 1 test
test credential_store::tests::write_private_file_follows_symlink_parent_directory ... ok
```

# Exploit Scenario
**Prerequisites**: Attacker has write access to `~/.config/` before the user first authenticates with file storage.
1. Attacker creates symlink: `ln -sf /tmp/world-readable ~/.config/s2`
2. User runs: `s2 login --insecure-storage` (or `s2 auth access-token set --insecure-storage`)
3. Credentials are written to the symlink target (e.g., `/tmp/world-readable/oauth-xxx.json`)
4. Later credential reads may fail due to read-path symlink checks, but the credential has already been leaked.

# Recommended Fix
Add symlink validation to `write_private_file()` (Unix) before and after `create_dir_all()` so the write path matches the read path’s safety checks (reject symlink parents / non-directories).

# History
This bug was introduced in commit cf6c2af. The commit implemented symlink validation on the read path (`secure_private_file_for_read`) but did not add the same validation on the write path (`write_private_file`), allowing credentials to be written through symlink parent directories.

infiniteregrets:
Fixes #701

The read path (`secure_private_file_for_read`) already refuses a symlinked parent directory via `symlink_metadata`, but the write path created directories, set permissions, and persisted the credential file through whatever the parent resolved to. A symlink planted at the credential directory (e.g. `~/.config/s2`) would silently redirect freshly written OAuth and access tokens elsewhere when using `--insecure-storage`.

`secure_directory` now performs the same non-following inspection as the read path before any bytes are written:

- Inspects the parent with `fs::symlink_metadata` (does not follow symlinks) and rejects anything that is not a real directory with `UnsafeCredentialFile`. This runs right after `create_dir_all` and before the temp file is created, so nothing — not even a temp file — is written through a symlink. It covers both existing and dangling symlinks, since `symlink_metadata` sees the link itself either way.
- Only tightens permissions to `0700` when group/other bits are actually set, matching the read path (previously it chmodded unconditionally, and through symlinks).

Verified against the repro in #701: the write now fails with `UnsafeCredentialFile` and the attacker directory stays empty. Existing `credential_store` tests pass.

Note: writes now fail for setups where `~/.config/s2` is a legitimately symlinked directory (e.g. dotfile managers) — but reads already failed there, so behavior is now consistent rather than newly restrictive.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1458,7 +1458,7 @@ version = "3.1.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "faf9468729b8cbcea668e36183cb69d317348c2e08e994829fb56ebfdfbaac34"
 dependencies = [
- "windows-sys 0.61.2",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
@@ -1885,7 +1885,7 @@ dependencies = [
  "libc",
  "option-ext",
  "redox_users",
- "windows-sys 0.61.2",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
@@ -2043,7 +2043,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "39cab71617ae0d63f51a36d69f866391735b51691dbda63cf6f96d042b63efeb"
 dependencies = [
  "libc",
- "windows-sys 0.61.2",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
@@ -2563,9 +2563,9 @@ dependencies = [
 
 [[package]]
 name = "h2"
-version = "0.4.15"
+version = "0.4.16"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6cb093c84e8bd9b188d4c4a8cb6579fc016968d14c99882163cd3ff402a4f155"
+checksum = "a9f37a958b41b3b19ee2707c06439c0e9e547e847223eb791ecb0cb821c65e27"
 dependencies = [
  "atomic-waker",
  "bytes",
@@ -3298,9 +3298,9 @@ checksum = "0ceec5bc11778974d1bcb055b18002eba7f4b3518b6a0081b3af5f21666da9ad"
 
 [[package]]
 name = "lru"
-version = "0.18.1"
+version = "0.18.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0b6180140927ee907000b0aa540091f6ea512ead4447c92b8fc35bc72788a5a6"
+checksum = "5d2f2f9b4ba7e6b24d95e7e899329d35be83bcded72c8540cdd5368932d1d90a"
 dependencies = [
  "hashbrown 0.17.1",
 ]
@@ -3530,7 +3530,7 @@ version = "0.50.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7957b9740744892f114936ab4a57b3f487491bbeafaf8083688b16841a4240e5"
 dependencies = [
- "windows-sys 0.61.2",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
@@ -4365,7 +4365,7 @@ dependencies = [
  "once_cell",
  "socket2",
  "tracing",
- "windows-sys 0.61.2",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


infiniteregrets:
@greptile-apps review

PR #720, opened 2026-09-06
feat: add Native storage class to shared interfaces

shikhar:
Add `native` to the shared API, SDK, resource specification, and CLI, using common ordinal 3. Standard (1), Express (2), and the Express default are preserved.

Use consistent append tail latency descriptions across the API and SDK: under 400 ms for Standard, 40 ms for Express, and 4 ms for Native.

Validation: all 792 tests in `just test`, full workspace Clippy, and `just fmt` pass.

This draft accompanies the [OpenAPI enum update](https://github.com/s2-streamstore/s2-specs/pull/22). Release both with consumer support for the new storage class.

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,6 +104,8 @@ headless host that has no credential store.
 
 It uses [SlateDB](https://slatedb.io) as its storage engine, which relies entirely on object storage for durability.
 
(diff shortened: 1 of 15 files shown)


shikhar:
moving to strings

PR #724, opened 2026-09-07
fix(lite): close SlateDB on graceful shutdown

chewbaucke:
Lite previously returned after draining HTTP requests without closing SlateDB, leaving memtables to be recovered from the WAL on the next startup.

Call `Backend::close()` after HTTP shutdown, delegating to SlateDB 0.15's `Db::close()` to flush memtables to L0 and shut down the database. Log the elapsed close time and propagate close errors. This reduces WAL replay on restart; Lite's existing `manifest_poll_interval` startup wait remains.

Shutdown awaits the database close without an application timeout. Deployments should allow enough termination grace for the final flush; an interrupted close may still leave WAL data to replay on the next startup.

Validation:

- `just fmt` and `just test` (744 tests passed).
- Local filesystem SIGTERM/reopen checks over HTTP and self-signed HTTPS preserved all 512 acknowledged records per scenario and accepted subsequent appends.
- The contributor reported restart-to-ready times of 2–6 seconds on Fly.io/Tigris, down from 106 seconds when quiet and 251–332 seconds after a burst, with a 1–2 second close and all 352 acknowledged records plus the seed record present after reopening. These remote-store timings were not independently reproduced during review.

### DIFF
--- a/lite/src/backend/core.rs
+++ b/lite/src/backend/core.rs
@@ -71,6 +71,13 @@ impl Backend {
         }
     }
 
+    /// Fold the memtable (L0) and close the writer. Planned stops must call
+    /// this after HTTP drain; SlateDB has no Drop close, so skipping it leaves
+    /// an un-folded WAL chain for the next fence/replay.
+    pub async fn close(&self) -> Result<(), slatedb::Error> {
+        self.db.close().await
+    }
+
     pub(super) fn bgtask_trigger(&self, trigger: BgtaskTrigger) {
         let _ = self.bgtask_trigger_tx.send(trigger);
     }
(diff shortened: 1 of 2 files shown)


shikhar:
thanks @chewbaucke makes sense!

chewbaucke:
> thanks @chewbaucke makes sense!

No problem! Discovered it while testing some changes on my own little [project](https://github.com/chewbaucke/spillbuf)

It's a tiny buffer to a persistent disk that sits in front of something that wants to publish to an s2 stream but ACK faster than it takes to upload it object storage. In my case it's for a webhook receiver so the sender can get a 200 right away.

PR #743, opened 2026-09-14
fix(sdk): finish advised reconnect after the final acknowledgement

DivyamTalwar:
## Summary

Addresses #722.

Finish an advised append reconnect once no submitted appends remain unacknowledged, instead of waiting for a trailing server EOF.

## Root cause and change

`run_session()` routes accepted reconnect advice to `drain_for_reconnect()`. That function half-closes the request sender and reads acknowledgements. `process_ack()` removes the acknowledged batch, resolves its ticket and disarms the acknowledgement timer when the queue empties. The previous drain loop nevertheless iterated again, armed a new timeout specifically to wait for response EOF, and could return `AckTimeout` even though every submitted record was acknowledged.

A pending response stream with an initially empty queue demonstrates the unnecessary wait. A second fixture yields the final valid acknowledgement and then remains pending: the queue and tickets are complete, but the old drain waits for EOF and can time out. The observed issue is unnecessary backoff/reconnect latency, **not demonstrated data loss**.

The patch makes the loop condition `!state.inflight_appends.is_empty()` and returns `Ok(())` after the queue drains. Half-close still occurs at entry. The existing response/error/deadline machinery remains in use while a batch is outstanding. No token refresh, retry policy, command submission, read session, public API, dependency, or server setting is changed.

## Verification actually performed

Eight repository-native Tokio regression tests are supplied, but Cargo/Rust are unavailable in the preparation environment. They have **not** been compiled or run. Do not mark this PR ready for review until native validation and formatting are complete.

Read `evidence/verification.json` and the test logs in the developer handoff. Local component results are not a substitute for the repository's required workflow. Update this section with the exact final native command/commit receipts when they exist; do not check unrun boxes.

## Scope, compatibility and remaining review

- Run and review the eight new Rust/Tokio tests; they have not been compiled here.
- Run the existing session/retry tests, workspace checks, and required formatter using the repository-approved toolchain.
- Refresh #722 and inspect pending work; do not revive the separate #721/#723 close-on-draining proposal.

AI-assisted candidate and single-assistant review. No independent reviewer consensus, full-suite success, live deployment or production certification is claimed. Only the listed source/test paths belong in this PR; the handoff, logs, inventory tools and audit infrastructure do not.


infiniteregrets:
hi @DivyamTalwar , i'd like to take a look into this issue myself! sorry

